Issue #11, opened 2024-02-07
Tracking issue for `Duration` component


nekevss:
Below is the functionality needed for the `Duration` component.

- [x] 7.3.3 get `Temporal.Duration.prototype.years`
- [x] 7.3.4 get `Temporal.Duration.prototype.months`
- [x] 7.3.5 get `Temporal.Duration.prototype.weeks`
- [x] 7.3.6 get `Temporal.Duration.prototype.days`
- [x] 7.3.7 get `Temporal.Duration.prototype.hours`
- [x] 7.3.8 get `Temporal.Duration.prototype.minutes`
- [x] 7.3.9 get `Temporal.Duration.prototype.seconds`
- [x] 7.3.10 get `Temporal.Duration.prototype.milliseconds`
- [x] 7.3.11 get `Temporal.Duration.prototype.microseconds`
- [x] 7.3.12 get `Temporal.Duration.prototype.nanoseconds`
- [x] 7.3.13 get `Temporal.Duration.prototype.sign`
- [x] 7.3.14 get `Temporal.Duration.prototype.blank` (implemented as `is_zero`)
- [x] 7.3.15 `Temporal.Duration.prototype.with ( temporalDurationLike )`
- [x] 7.3.16 `Temporal.Duration.prototype.negated ( )`
- [x] 7.3.17 `Temporal.Duration.prototype.abs ( )`
- [x] 7.3.18 `Temporal.Duration.prototype.add ( other [ , options ] )`
- [x] 7.3.19 `Temporal.Duration.prototype.subtract ( other [ , options ] )`
- [x] 7.3.20 `Temporal.Duration.prototype.round ( roundTo )`
- [x] 7.3.21 `Temporal.Duration.prototype.total ( totalOf )`
- [x] 7.3.22 `Temporal.Duration.prototype.toString ( [ options ] )`
- [x] 7.3.23 `Temporal.Duration.prototype.toJSON ( )`
- [x] 7.3.24 `Temporal.Duration.prototype.toLocaleString ( [ locales [ , options ] ] )` (not yet implementable - needs `icu_duration` if it were to exist)
- [x] 7.3.25 `Temporal.Duration.prototype.valueOf ( )`

nekevss:
This has long been "done", but I've left open for a couple reasons:

1. the inner representation of Duration can still be optimized
2. there is a known precision bug on maximum milliseconds, microseconds, and nanoseconds.

[...]

PR #17, opened 2024-02-08
Add blank and negated + small adjustments

nekevss:
Related to #11

Primarily implementing `blank` and `negated` functionality onto blank.

Renames: `duration_sign` -> `sign`

### DIFF
--- a/src/components/duration.rs
+++ b/src/components/duration.rs
@@ -15,7 +15,6 @@ mod time;
 
 #[doc(inline)]
 pub use date::DateDuration;
-use num_traits::Zero;
 #[doc(inline)]
 pub use time::TimeDuration;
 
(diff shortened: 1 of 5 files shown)


src/components/duration.rs
```diff
@@ -857,23 +857,36 @@ impl Duration {
...
     #[must_use]
-    pub fn abs(&self) -> Self {
+    pub fn sign(&self) -> i32 {
+        duration_sign(&self.iter().collect())
+    }
+
+    /// Returns whether the current `Duration` is blank.
+    pub fn blank(&self) -> bool {
+        duration_sign(&self.iter().collect()) == 0
+    }
```
jedel1043, line 870: Would probably be better to name this `is_zero`, I think.

jedel1043: I was more talking about the function itself, since `duration_sign` returns zero only when `self` is zero.

nekevss: Oh, I'm blind lol yeah, that can work too.

jedel1043 (approved):
LGTM!

src/components/duration.rs
nekevss: Glancing back through old temporal meeting minutes. It used to be called `isZero` but was opted off of, but I see no reason to not go with `isZero` as long as the equivalency is noted.

jedel1043: Yeah, `blank` is not very descriptive IMHO. We can add some comments when we rename spec methods just like in Boa.

Issue #19, opened 2024-02-09
Implement `Duration` normalization and new algorithms


nekevss:
The proposal recently updated `Duration` in tc39/proposal-temporal#2722, tc39/proposal-temporal#2727, and tc39/proposal-temporal#2612. This will need to be updated accordingly.

EDIT: There's been an overhaul of the implementation of the duration methods in general on top of normalization that needs to be updated.

PR #20, opened 2024-02-10
Implement `Duration` normalization - Part 1

nekevss:
This draft PR is meant to eventually address #19

Currently, this PR is small baseline changes focused on `NormativeTime` records and `balance` methods (EDIT: it also adds `CalendarDateAdd` functionality for the `ISO8601`). ~~The changes overall are pretty widespread across large sections of `Duration` that could really use more tests, so the plan is to come back to this once it can be tested a bit more or split it into a couple of PRs.~~

I'm going to switch this into a normal PR from a draft. After looking through it, #19 will still need to be implemented in a couple PRs, but I think it makes more sense to break the overall update down into smaller chunks

### DIFF
--- a/src/components/duration/time.rs
+++ b/src/components/duration/time.rs
@@ -5,7 +5,7 @@ use crate::{
     utils, TemporalError, TemporalResult,
 };
 
-use super::is_valid_duration;
+use super::{is_valid_duration, normalized::NormalizedTimeDuration};
 
 /// `TimeDuration` represents the [Time Duration record][spec] of the `Duration.`
 ///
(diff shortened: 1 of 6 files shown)


jedel1043:
I'm seeing the spec corrections make it possible to do the computations using 128bit integers. Will we transition to that representation in the future? (32 bit components + 128 bit normalized repr)

nekevss:
Yeah, I think it did.

To be frank, I haven't put too much thought into it as of yet. Mostly because I would really like to have a lot more of the spec implemented and some more tests in place to mitigate any edge cases or oddities that may arise off the switch.

Since `f128` isn't implemented yet in Rust maybe that should be prioritized after the normalization updates.

src/components/duration/time.rs
```diff
@@ -418,11 +397,17 @@ impl TimeDuration {
     pub fn iter(&self) -> TimeIter<'_> {
         <&Self as IntoIterator>::into_iter(self)
     }
+
+    /// Returns this `TimeDuration` as a `NormalizedTimeDuration`.
+    pub(crate) fn as_norm(&self) -> NormalizedTimeDuration {
```
jedel1043, line 402:
```suggestion
    pub(crate) fn to_normalized(&self) -> NormalizedTimeDuration {
```

`norm` has a different meaning in math contexts, and the `as_` prefix is commonly used to do simple casts.

jedel1043 (approved):
Pretty straightforward changes, mostly because it's closely following the spec. I just have a single suggestion that doesn't block merging this.

jedel1043 (approved):

PR #23, opened 2024-02-14
Duration Normalization - Part 2

nekevss:
This PR is a continuation on #20 implementing the `Duration` normalization changes. I think this gets close to covering most a decent portion of the changes major changes. This also shifts the `round_to_increment` functions to `i64`.


jedel1043 (approved):
Most of the comments I have are already marked with TODOs, so we can merge this as it is.

PR #24, opened 2024-02-17
Implement `Duration::round` and some general updates/fixes

nekevss:
~~Currently a WIP.~~ 
EDIT: this is ready for review!

This PR is to implement `Duration::round` along with some other updates and ideally basic tests as well. The hope is to have this implemented and tested so that it can be used to switch over the internal duration slots from `f64` to `i32`. 😄

A couple notes:
1. I ended up switching rounding back to f64 for now / this PR. It had been u64, and my goal for the next PR is to head back in that direction. The truncation of values between integer types was causing some havoc and making debugging more difficult than it needed to be while dealing with `round`.
2. Currently, the only testing is for `floor` rounding, but this can be added to if desired.


src/components/duration/date.rs
```diff
@@ -38,6 +38,292 @@ impl DateDuration {
...
+        // 1. Assert: If plainRelativeTo is not undefined, calendarRec is not undefined.
+        let plain_relative = plain_relative_to.expect("plainRelativeTo must not be undefined.");
+
+        // 2. Let defaultLargestUnit be DefaultTemporalLargestUnit(years, months, weeks, days, 0, 0, 0, 0, 0).
+        let default_largest = self
+            .iter()
+            .enumerate()
+            .find(|x| x.1 != 0.0)
+            .map(|x| TemporalUnit::from(10 - x.0))
+            .unwrap_or(TemporalUnit::Nanosecond);
```
jedel1043, line 58: Seems weird to have `Duration` implement `IntoIterator` only for this. Maybe we can just do a manual check of all its components instead?

nekevss: The `IntoIterator` implementation is also used to collect the fields and check the duration's sign and it's validity. This could be refactored to manually check the fields instead of the `IntoIterator` implementation, but I figured using an iterator would be a bit more ergonomic.

That being said, if we were to keep the `IntoIterator` impl, I'm currently wondering whether it should be public.

jedel1043: Yeah, the problem is more that we're exposing implementations that are only used internally; it doesn't make much sense to implement IntoIterator for a Duration type. You could also extract that into a function, but I figured it would be easier to just have a `largest_unit` method instead.

nekevss: Removed the `IntoIterator` implementations in favor of an internal `fields` method to return the fields in a `Vec` and did a little clean up to the API along with it to pull back a bit on what we'd be exposing.

jedel1043 (approved):
Looks good! Let's merge this and publish the first version of `temporal_rs`.

PR #25, opened 2024-02-23
Add `non_exhaustive` attribute to component structs

nekevss:
Updates public facing structs to have a `#[non_exhaustive]` attribute flag so that components cannot be constructed with a `StructExpression`.


jasonwilliams (approved):
LGTM

PR #26, opened 2024-02-25
Setup publish CI

jedel1043:
Maybe we could think about using [release-plz](https://release-plz.ieni.dev/) in the future, but this should suffice for now.


cliff.toml
```diff
@@ -0,0 +1,83 @@
...
+<!-- generated by git-cliff -->
+"""
+# postprocessors
+postprocessors = []
+
+[git]
+# parse the commits based on https://www.conventionalcommits.org
+conventional_commits = false
+# filter out the commits that are not conventional
+filter_unconventional = true
```
nekevss, line 62: question: This should probably be false for now, right?

So far commits have not used conventional commits (although, maybe we should use them but that's a different discussion), and this filters any commits that have not used conventional commits, which would be everything.

nekevss (commented):
Looks good 😄 Just had one question.

cliff.toml
jedel1043: True! I don't know why cliff generates the default template like this (I picked the Github template), but I'll set this to false just in case.

nekevss (approved):

PR #28, opened 2024-02-25
Remove keywords from Cargo.toml

jedel1043:
```
error: failed to publish to registry at https://crates.io/
Caused by:
  the remote server responded with an error: expected at most 5 keywords per crate
```


nekevss (approved):
Oops, that's my bad. 😅 LGTM

jedel1043:
Also noticed that the categories don't exist on crates.io. Will change that just in case.

nekevss:
Looking at generally similar libraries, "date", "time", "calendar", "timezone", and "duration" might be best.

[Lib.rs](https://lib.rs/search?q=date)

jedel1043:
Yeah, we can publish using those and modify them if needed in the future.

nekevss (approved):

PR #30, opened 2024-02-29
Small parser refactor / cleanup

nekevss:
This PR streamlines some of the ISO8601 parsing and removes unnecessary iterations.

### DIFF
--- a/src/parser/time.rs
+++ b/src/parser/time.rs
@@ -66,16 +66,11 @@ pub(crate) fn parse_time_spec(cursor: &mut Cursor) -> TemporalResult<TimeSpec> {
     })
 }
 
+/// Parse an hour value.
+#[inline]
 pub(crate) fn parse_hour(cursor: &mut Cursor) -> TemporalResult<u8> {
-    let start = cursor.pos();
-    for _ in 0..2 {
-        let digit = cursor.abrupt_next()?;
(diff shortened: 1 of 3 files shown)


src/parser/time.rs
```diff
@@ -66,35 +66,23 @@ pub(crate) fn parse_time_spec(cursor: &mut Cursor) -> TemporalResult<TimeSpec> {
...
+    let hour_value = first * 10 + cursor.next_digit()?;
     if !(0..=23).contains(&hour_value) {
         return Err(TemporalError::syntax().with_message("Hour must be in a range of 0-23"));
     }
     Ok(hour_value)
 }
 
 // NOTE: `TimeSecond` is a 60 inclusive `MinuteSecond`.
-/// Parse `MinuteSecond`
+/// Parse `MinuteSecond` value.
```
jedel1043, line 81: Can you link to the corresponding grammar rule?

jedel1043 (approved):
Nice refactor! Makes it very clean for numeric parses.

src/parser/time.rs
nekevss: Can do. Better documentation all the way around is definitely needed. Should we update the grammar rules as a whole in a follow-up PR or do you mind if I add them in on this PR?

jedel1043: If you're planning on doing a PR only for documentation, you can move this change to that PR.

PR #31, opened 2024-03-03
Fix loop in `diff_iso_date`

### LINKED ISSUE #29: Bug: infinite loop in `CalendarDateUntil` calculation
While migrating Boa to `temporal_rs` in boa-dev/boa#3694, an infinite loop was discovered in the `calendar.prototype.dateUntil` test suite.

Cause: most likely occurs [here](https://github.com/boa-dev/temporal/blob/main/src/iso.rs#L243) in the `diff_iso_date` function.

Test 262 tests: The loop occurs in the [largest-unit-year.js](https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Calendar/prototype/dateUntil/largest-unit-year.js) tests.

nekevss:
Closes #29

Fixes the issue that was being causing `diff_iso_date` to loop and notes it for the future.

Adds the test case that was causing the initial issue.


nekevss:
There's an issue with calculating negative durations. I'm working on updating the above.

src/iso.rs
```diff
@@ -733,8 +735,12 @@ fn iso_date_surpasses(this: &IsoDate, other: &IsoDate, sign: i8) -> bool {
 
 #[inline]
 fn balance_iso_year_month(year: i32, month: i32) -> (i32, i32) {
-    let y = year + (month - 1) / 12;
+    // 1. Assert: year and month are integers.
+    // 2. Set year to year + floor((month - 1) / 12).
+    let y = year + (month - 1).div_euclid(12);
```
jedel1043, line 740: Oof, that's a really tricky bug.

jedel1043 (approved):
Nice catch!

src/iso.rs
nekevss: It took longer than I'd like to admit to notice it too 😭

PR #34, opened 2024-03-10
Fix panics when no relative_to is supplied to round

robot-head:


src/components/duration.rs
```diff
@@ -568,21 +568,23 @@ impl Duration {
         };
         // 35. Let calendarRec be ? CreateCalendarMethodsRecordFromRelativeTo(plainRelativeTo, zonedRelativeTo, « DATE-ADD, DATE-UNTIL »).
 
-        // TODO: Validate below expect -> Potentially return error.
-        let relative_to_date = relative_to.date.expect("Date must exist at this point.");
```
nekevss, line 572: suggestion: This can actually just be:

```rust

let relative_to_date = relative_to.date;

```

You should be able to just provide `relative_to_date` as an arg instead of `Some(relative_to_date)`

nekevss (commented):
Nice catch on this!

I have one suggestion on how to do this without overly changing the structure, which I believe should work. There's definitely a need to validate whether the function calls that would now be provided a `None` are properly handling the case.

PR #35, opened 2024-03-16
Implements `Date`'s `add`, `subtract`, `until`, and `since` methods

nekevss:
Implements `add`, `subtract`, `until`, and `since` methods for the `Date` component with basic tests for all methods.

This also updates the visibility for some of the abstract methods and `NormalizedDurationRecord`.

### DIFF
--- a/src/components/date.rs
+++ b/src/components/date.rs
@@ -9,9 +9,9 @@ use crate::{
         DateTime, Duration,
     },
     iso::{IsoDate, IsoDateSlots},
-    options::{ArithmeticOverflow, TemporalUnit},
+    options::{ArithmeticOverflow, RelativeTo, TemporalRoundingMode, TemporalUnit},
     parser::parse_date_time,
-    TemporalError, TemporalResult,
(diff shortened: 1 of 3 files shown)


src/components/date.rs
```diff
@@ -45,11 +45,185 @@ impl<C: CalendarProtocol> Date<C> {
...
+    pub(crate) fn diff_date(
+        &self,
+        op: bool,
+        other: &Self,
+        rounding_mode: Option<TemporalRoundingMode>,
+        rounding_increment: Option<f64>,
+        largest_unit: Option<TemporalUnit>,
+        smallest_unit: Option<TemporalUnit>,
+        context: &mut C::Context,
+    ) -> TemporalResult<Duration> {
```
jedel1043, line 131: Same recommendation as the other PR. It would be better to do the sign calculation and application outside of this function, and the rounding type can just be passed as an argument.

jedel1043 (commented):
Good work! I just have a small comment.

jedel1043 (approved):

PR #36, opened 2024-03-17
Implement Time's until and since methods

nekevss:
Implements the `until` and `since` methods for the `Time` component.

### DIFF
--- a/src/components/time.rs
+++ b/src/components/time.rs
@@ -19,6 +19,7 @@ pub struct Time {
 impl Time {
     #[inline]
     #[must_use]
+    /// Creates a new unvalidated `Time`.
     pub(crate) fn new_unchecked(iso: IsoTime) -> Self {
         Self { iso }
     }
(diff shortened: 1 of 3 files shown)


src/components/time.rs
```diff
@@ -46,6 +47,69 @@ impl Time {
...
+            0.0,
+            0.0,
+            sign * result.hours,
+            sign * result.minutes,
+            sign * result.seconds,
+            sign * result.milliseconds,
+            sign * result.microseconds,
+            sign * result.nanoseconds,
+        )
+    }
```
jedel1043, line 112: I think in this case it would be better to extract the sign and rounding type calculation outside of the function, and just do that on the specific `until` and `since` methods. It would make this a lot cleaner.

jedel1043 (commented):
Everything looks good! just have a small suggestion.

src/components/time.rs
nekevss: Yeah, I was actually going to post that as a potential issue for discussion soon regarding handling those options. I think that entire section needs to be probably discussed as to how to handle it.

Right now the options parameters are separate and there is a `RelativeTo` parameter on `round` functions. I wasn't the biggest fan of that structure, and after playing around with it and review the recent PRs involving it. I think it would be better to remove `RelativeTo`, but at the same time, to still simplify the function parameters, add a `DifferenceSettings` struct. All the lines from 65-79 could then be extracted into a method on `DifferenceSettings` something like `difference_settings.prepare()`.

But this is mostly initial ideation on my part. Any thoughts?

jedel1043: That would be pretty cool! Would also allow extending the function with more options in the future without a breaking change, which is really nice :)

jedel1043: Since that change is on the plans then, I think you can ignore this review and do the changes on a separate PR.

jedel1043 (approved):

PR #41, opened 2024-04-16
Extract temporal tests from the test262 suite

jedel1043:
Just a POC for now, and only generates tests for `Duration::add`, but theoretically with the correct glue code we could transform all Temporal tests into a better format for us.

Still need to setup the workspace for it.

### DIFF
--- a/test262_config.toml
+++ b/test262_config.toml
@@ -1,0 +1,1 @@
+commit = "6f7ae1f311a7b01ef2358de7f4f6fd42c3ae3839"
--- a/testgen/Cargo.toml
+++ b/testgen/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "temporal-tests"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+boa_engine = "0.18.0"
+boa_gc = "0.18.0"
+color-eyre = "0.6.2"
+rustc-hash = "1.1.0"
+serde = { version = "1.0.197", features = ["derive"] }
+serde_repr = "0.1.18"
+serde_yaml = "0.9.32"
+bitflags = "2.4.2"
+regex = "1.10.3"
+phf = "0.11.2"
+rayon = "1.9.0"
+log = "0.4.21"
+simple_logger = "4.3.3"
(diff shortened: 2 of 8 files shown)


testgen/src/exec/temporal.rs
```diff
@@ -0,0 +1,158 @@
...
+    let years: i32 = args.get_or_undefined(1).try_js_into(context)?;
+    let months: i32 = args.get_or_undefined(2).try_js_into(context)?;
+    let weeks: i32 = args.get_or_undefined(3).try_js_into(context)?;
+    let days: i32 = args.get_or_undefined(4).try_js_into(context)?;
+    let hours: i32 = args.get_or_undefined(5).try_js_into(context)?;
+    let minutes: i32 = args.get_or_undefined(6).try_js_into(context)?;
+    let seconds: i32 = args.get_or_undefined(7).try_js_into(context)?;
+    let milliseconds: i32 = args.get_or_undefined(8).try_js_into(context)?;
+    let microseconds: i32 = args.get_or_undefined(8).try_js_into(context)?;
+    let nanoseconds: i32 = args.get_or_undefined(9).try_js_into(context)?;
```
HalidOdat, line 135: Same with `to_integer_if_integral` the `TryFromJs` trait for `i32` [doesn't seem to be checking for floating point values that are integers](https://github.com/boa-dev/boa/blob/main/core/engine/src/value/conversions/try_from_js.rs#L216-L225).

This seem like an unrelated bug in `TryFromJs` that should be solved by https://github.com/boa-dev/boa/issues/3761, but maybe we should account for it here, it maybe causing some test to not generate?

testgen/src/read.rs
```diff
@@ -0,0 +1,214 @@
...
+        })
+        .captures(code.as_bytes())
+        .ok_or_eyre("missing metadata for test")?
+        .get(1)
+        .map(|m| String::from_utf8_lossy(m.as_bytes()))
+        .ok_or_eyre("invalid metadata for test")?
+        .replace('\r', "\n");
+
+    serde_yaml::from_str(&yaml).map_err(Into::into)
+}
```
HalidOdat, line 214: I think we have a better version on `main`: https://github.com/boa-dev/boa/blob/main/tests/tester/src/read.rs#L218-L231

We may want to extract the tester to it's own crate at some point, it's easier now that we have the test262 running nightly, to avoid having so much duplicate code :sweat_smile:

HalidOdat (commented):
Looks awesome! Have some comments, see if they make sense :)

nekevss:
Closing this PR as it's fairly out of date. Would still be nice to test directly against test262. But we have enough proxies I think between V8, Boa, Kiesel, and their runtimes.

PR #44, opened 2024-05-17
adding methods for yearMonth and MonthDay

jasonwilliams:
Adds methods to yearMonth and monthDay

### DIFF
--- a/src/components/month_day.rs
+++ b/src/components/month_day.rs
@@ -42,20 +42,6 @@ impl MonthDay {
         Ok(Self::new_unchecked(iso, calendar))
     }
 
-    /// Returns the `month` value of `MonthDay`.
-    #[inline]
-    #[must_use]
-    pub fn month(&self) -> u8 {
-        self.iso.month
-    }
-
-    /// Returns the `year` value of `MonthDay`.
-    #[inline]
-    #[must_use]
-    pub fn year(&self) -> i32 {
-        self.iso.year
-    }
-
     /// Returns the `day` value of `MonthDay`.
     #[inline]
     #[must_use]
@@ -72,7 +58,6 @@ impl MonthDay {
 
     /// Returns the `monthCode` value of `MonthDay`.
     #[inline]
-    #[must_use]
     pub fn month_code(&self) -> TemporalResult<TinyAsciiStr<4>> {
         self.calendar
             .month_code(&CalendarDateLike::MonthDay(self.clone()))
--- a/src/components/month_day.rs
+++ b/src/components/month_day.rs
@@ -11,7 +11,7 @@ use crate::{
     TemporalError, TemporalResult, TemporalUnwrap,
 };
 
-use super::calendar::GetTemporalCalendar;
+use super::calendar::{CalendarDateLike, GetTemporalCalendar};
 
 /// The native Rust implementation of `Temporal.PlainMonthDay`
 #[non_exhaustive]
@@ -69,25 +69,20 @@ impl MonthDay {
     pub fn calendar(&self) -> &Calendar {
         &self.calendar
     }
+
+    /// Returns the `monthCode` value of `MonthDay`.
+    #[inline]
+    #[must_use]
+    pub fn month_code(&self) -> TemporalResult<TinyAsciiStr<4>> {
+        self.calendar
+            .month_code(&CalendarDateLike::MonthDay(self.clone()))
+    }
 }
 
 impl GetTemporalCalendar for MonthDay {
     fn get_calendar(&self) -> Calendar {
         self.calendar.clone()
     }
-
-// Contextual Methods
-impl<C: CalendarProtocol> MonthDay<C> {
-    pub fn contextual_month_code(
-        this: &C::MonthDay,
-        context: &mut C::Context,
-    ) -> TemporalResult<TinyAsciiStr<4>> {
-        this.get_calendar()
-            .month_code(&CalendarDateLike::MonthDay(this.clone()), context)
-    }
-}
-
-
 }
 
 impl IsoDateSlots for MonthDay {
--- a/src/components/month_day.rs
+++ b/src/components/month_day.rs
@@ -42,17 +42,17 @@ impl MonthDay {
         Ok(Self::new_unchecked(iso, calendar))
     }
 
-    /// Returns the `day` value of `MonthDay`.
+    /// Returns the iso day value of `MonthDay`.
     #[inline]
     #[must_use]
-    pub fn day(&self) -> u8 {
+    pub fn iso_day(&self) -> u8 {
         self.iso.day
     }
 
-    // returns the `month` value of `MonthDay`.
+    // returns the iso month value of `MonthDay`.
     #[inline]
     #[must_use]
-    pub fn month(&self) -> u8 {
+    pub fn iso_month(&self) -> u8 {
         self.iso.month
     }
 
--- a/src/components/month_day.rs
+++ b/src/components/month_day.rs
@@ -49,6 +49,20 @@ impl MonthDay {
         self.iso.day
     }
 
+    // returns the `month` value of `MonthDay`.
+    #[inline]
+    #[must_use]
+    pub fn month(&self) -> u8 {
+        self.iso.month
+    }
+
+    /// Returns the string identifier for the current calendar used.
+    #[inline]
+    #[must_use]
+    pub fn calendar_id(&self) -> String {
+        self.calendar.identifier()
+    }
+
     /// Returns a reference to `MonthDay`'s `CalendarSlot`
     #[inline]
     #[must_use]
--- a/src/components/month_day.rs
+++ b/src/components/month_day.rs
@@ -2,6 +2,8 @@
 
 use std::str::FromStr;
 
+use tinystr::TinyAsciiStr;
+
 use crate::{
     components::calendar::Calendar,
     iso::{IsoDate, IsoDateSlots},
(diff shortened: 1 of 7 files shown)


nekevss (approved):
Looks good to me whenever you want to take it off draft.

The one thing missing is maybe the default implementation and some unit tests, but I don't think those are necessarily needed to merge :smile:

jasonwilliams:
@nekevss let me know your thoughts

I've been trying to add add() but it seems it's iso is using the [date version](https://github.com/boa-dev/temporal/blob/dd51ebbf92fb48701c3236af0c3fb302ea180e30/src/iso.rs#L243-L276) of add_date_duration, but it seems like i need an implementation more like dateTime's: 
https://github.com/boa-dev/temporal/blob/dd51ebbf92fb48701c3236af0c3fb302ea180e30/src/iso.rs#L125-L164

The former doesn't use the calendar at all by the looks of it

secondly why does the leap_year not return a bool? Is that wasteful?
https://github.com/boa-dev/temporal/blob/dd51ebbf92fb48701c3236af0c3fb302ea180e30/src/utils.rs#L230-L233

src/components/year_month.rs
```diff
@@ -61,6 +66,62 @@ impl<C: CalendarProtocol> YearMonth<C> {
...
+        this.get_calendar()
+            .days_in_year(&CalendarDateLike::YearMonth(this.clone()), context)
+    }
+
+    pub fn get_days_in_month(this: &C::YearMonth, context: &mut C::Context) -> TemporalResult<u16> {
+        this.get_calendar()
+            .days_in_month(&CalendarDateLike::YearMonth(this.clone()), context)
+    }
+
+    fn add_or_subtract_duration(
```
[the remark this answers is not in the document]
jasonwilliams: true
I think `self.iso_date()` object will need a add_duration method, as right now it only seems to support adding date_durations but it should be able to support any type of durations really.

src/fields.rs
```diff
@@ -295,7 +295,7 @@ impl TemporalFields {
         let FieldValue::Integer(mo) = value else {
             return Err(TemporalError::r#type().with_message("Month must be an integer."));
         };
-        self.year = Some(*mo);
+        self.month = Some(*mo);
```
nekevss, line 298: Nice catch!

src/components/year_month.rs
```diff
@@ -61,6 +68,80 @@ impl<C: CalendarProtocol> YearMonth<C> {
...
+
+    pub fn contextual_get_months_in_year(
+        this: &C::YearMonth,
+        context: &mut C::Context,
+    ) -> TemporalResult<u16> {
+        this.get_calendar()
+            .months_in_year(&CalendarDateLike::YearMonth(this.clone()), context)
+    }
+
+    pub fn contextual_add_or_subtract_duration(
```
nekevss, line 114: issue: this method should probably be `pub(crate)` at most.

`AddOrSubtractDuration` should probably be an internal abstract method. I think ideally, the public API here should be `add` & `subtract`. This would also save adding the `DurationOperation` enum.

src/components/year_month.rs
```diff
@@ -61,6 +68,80 @@ impl<C: CalendarProtocol> YearMonth<C> {
...
+        mut duration: Duration,
+        context: &mut C::Context,
+        overflow: ArithmeticOverflow,
+    ) -> TemporalResult<YearMonth<C>> {
+        match operation {
+            DurationOperation::Subtract => duration = duration.negated(),
+            DurationOperation::Add => {}
+        }
+
+        let mut fields = TemporalFields::default();
```
nekevss, line 126: thought: huh, seeing this makes me wonder if it would be worthwhile to implement a  `From<YearMonth> for TemporalFields` and `From<Date> for TemporalFields`

jasonwilliams: its difficult to do this as this is an associated type `From<&<C as CalendarProtocol>::YearMonth>`, i couldn't find a way to make this work

nekevss: Follow up! So I think the best approach would be below. Found it when working the new `Date` methods. :smile: 

```rust

let fields = TemporalFields {
  bit_map: FieldMap::YEAR | FieldMap::MONTH,
  year: Some(this.iso_date().year),
  month: Some(this.iso_date().month),
  ..Default::default(),
}

```
Also, there may be an argument for implementing the from as a tuple, `impl From<(i32, u8)> for TemporalFields`. That might be able to be extrapolated out a bit further with a wrapper to ensure it's being converted properly (granted this also may be doing too much). 

```rust

struct YearMonthFields((i32, u8));

impl From<YearMonthFields> for TemporalFields {
  fn from(value: YearMonthFields) -> Self {
    TemporalFields {
      bit_map: FieldMap::YEAR | FieldMap::MONTH,
      year: Some(value.0.0),
      month: Some(value.0.1),
      ..Default::default(),
    }
  }
}

```

`MonthDay` could then be very similar with (u8, u8) in that instance. Let me know what you think!

jasonwilliams: Yep I like this idea, looks great, thank you

src/components/duration.rs
```diff
@@ -24,6 +24,11 @@ pub use date::DateDuration;
 #[doc(inline)]
 pub use time::TimeDuration;
 
+#[derive(Debug, Clone, Copy)]
+pub enum DurationOperation {
```
nekevss, line 28: Is `DurationOperation` required anymore?

src/components/year_month.rs
```diff
@@ -2,14 +2,24 @@
...
+};
+
+use super::{
+    calendar::{CalendarDateLike, GetTemporalCalendar},
+    Duration,
 };
 
-use super::calendar::GetTemporalCalendar;
+// Subset of `TemporalFields` representing just the  `YearMonthFields`
+pub struct YearMonthFields(i32, u8);
```
nekevss, line 22: question: is there a primary use case for this?

I think it's mostly used to implement `impl From<YearMonthFields> for TemporalFields`, and I'm sort of curious if it would just be better to implement `From<YearMonth>`

nekevss (changes requested):

src/components/year_month.rs
jasonwilliams: Yes I think so, I can take a look to see if we can just implement From

jasonwilliams: This is used for add_or_subtract_duration to get the fields of YearMonth then generate a Date object

src/components/month_day.rs
```diff
@@ -35,22 +37,30 @@ impl MonthDay {
...
+        // 1972 is the first leap year in the Unix epoch (needed to cover all dates)
         let iso = IsoDate::new(1972, month, day, overflow)?;
         Ok(Self::new_unchecked(iso, calendar))
     }
 
-    /// Returns the `month` value of `MonthDay`.
+    /// Returns the `day` value of `MonthDay`.
+    #[inline]
+    #[must_use]
+    pub fn day(&self) -> u8 {
```
nekevss, line 48: nit (non-blocking): this should probably be `iso_day`.

src/components/year_month.rs
```diff
@@ -48,19 +57,97 @@ impl YearMonth {
         self.iso.year
     }
 
+    /// Returns the padded ISO year string
+    #[inline]
+    #[must_use]
+    pub fn padded_iso_year_string(&self) -> String {
```
nekevss, line 63: thought: I'm not entirely sure long term this method should be in our public API.

This method is used for the `toString` implementation on the boa side I believe, which I would much prefer to have the entire `to_string` method implemented on `YearMonth` rather than component parts.

That being said, it's underlying utility method looks pretty fantastic, so I'm not against keeping it and iterating on it until there's a full `to_string` implementation in `temporal_rs`

src/components/year_month.rs
```diff
@@ -48,19 +57,97 @@ impl YearMonth {
...
 
+    pub fn month_code(&self) -> TemporalResult<TinyAsciiStr<4>> {
+        self.get_calendar()
+            .month_code(&CalendarDateLike::YearMonth(self.clone()))
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn in_leap_year(&self) -> bool {
+        utils::mathematical_in_leap_year(utils::epoch_time_for_year(self.iso.year)) == 1
```
nekevss, line 82: issue: this should probably be a call to the `Calendar`'s in leap year method.

```
self.get_calendar().in_leap_year(&CalendarDateLike::YearMonth(self.clone()))
```

nekevss (commented):
There's one last thing that I missed the last couple times around. But outside of that this is overall looking fantastic!

src/components/year_month.rs
jasonwilliams: Yep ok, we can come back to this but ill leave it here for now and we can decide going forward  how we want to handle to_string methods

nekevss (approved):
Looks great!

jedel1043:
Oh, I had an unsent pending review 😅 I'll push the comments and I think we can address them on a separate PR

src/components/year_month.rs
```diff
@@ -2,14 +2,23 @@
...
 };
 
-use super::calendar::GetTemporalCalendar;
+use super::{
+    calendar::{CalendarDateLike, GetTemporalCalendar},
+    Duration,
+};
+
+// Subset of `TemporalFields` representing just the  `YearMonthFields`
+pub struct YearMonthFields(pub i32, pub u8);
```
jedel1043, line 21: Why do we need to expose the fields as public?
Also, I think this should be a proper struct with named fields for clarity

PR #48, opened 2024-06-02
Patch `(un)balance_relative` to avoid panicking

jedel1043:
A simple patch to avoid panics in Boa, since we'll eventually fix that when the spec is a bit more stable.

### DIFF
--- a/src/components/duration/date.rs
+++ b/src/components/duration/date.rs
@@ -8,7 +8,7 @@ use crate::{
     options::{
         ArithmeticOverflow, RelativeTo, RoundingIncrement, TemporalRoundingMode, TemporalUnit,
     },
-    utils, TemporalError, TemporalResult, NS_PER_DAY,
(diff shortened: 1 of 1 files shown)


nekevss (approved):
Good patch. Definitely need to avoid using expect and use some sort of assertion error :face_with_diagonal_mouth:

jedel1043:
> Good patch. Definitely need to avoid using expect and use some sort of assertion error :face_with_diagonal_mouth: 

Yeah, it would be nice to have the library be panic free.

nekevss:
Yeah, I was already adding an assertion error into the parser PR. Maybe it would be better off as a separate PR but it should be soon. I'm definitely not super thrilled with using `expect` on asserts without a good way to test everything.

nekevss:
Is it worth rebasing and using the new functionality from #50

jedel1043:
Yep, gonna do that

PR #49, opened 2024-06-02
Add struct to encapsulate invariants of rounding increments

jedel1043:
This saves a bit of memory on the parameter list (`Option<NonZeroU32>` is 4 bytes, while `Option<f64>` is 16 bytes).


src/options.rs
```diff
@@ -3,13 +3,114 @@
...
+        // 3. If increment > maximum, throw a RangeError exception.
+        if increment > max {
+            return Err(TemporalError::range().with_message("roundingIncrement exceeds maximum."));
+        }
+
+        // 4. If dividend modulo increment ≠ 0, then
+        if dividend.rem_euclid(increment) != 0 {
+            //     a. Throw a RangeError exception.
+            return Err(TemporalError::range()
+                .with_message("dividend is not divisble by roundingIncrement."));
```
nekevss, line 106: nit: divisible

nekevss (approved):
One nit, but definitely always down for a better repr :smile:

nekevss:
Looks like this needs a rebase

PR #50, opened 2024-06-02
Migrate parsing to `ixdtf` crate

### LINKED ISSUE #39: Migrate to the `ixdtf` crate for datetime parsing
Right now the [`ixdtf`](https://crates.io/crates/ixdtf) crate is just a placeholder, but @nekevss migrated the parser to the ICU4X repo on https://github.com/unicode-org/icu4x/pull/4646. We should migrate to that when the next version of `ixdtf` releases.

nekevss:
This would close #39.

This migrates us from the current parser to the `ixdtf` crate.

As mentioned in #48, this also adds a macro `temporal_assertion` for asserting with the goal of being able to use that functionality instead of `expect` going forward.


src/lib.rs
```diff
@@ -60,6 +60,16 @@ pub use fields::TemporalFields;
...
+#[macro_export]
+macro_rules! temporal_assertion {
+    ($assertion:expr) => {{
+        let Some(value) = $assertion else {
+            return Err(TemporalError::assert());
+        };
+        value
+    }};
+}
+
```
jedel1043, line 72: This makes deconstruction a bit more opaque, which I don't prefer.

What about a `TemporalUnwrap` trait implemented for `Option<T>? It could be interesting that way, since we could extend it in the future with a `context` parameter if we need to.

jedel1043 (commented):
Overall good, I just want to have a discussion on the new macro.

src/lib.rs
nekevss: Oooh, that's a nice idea. Give me a few minutes to put it together.

src/lib.rs
```diff
@@ -60,6 +60,20 @@ pub use fields::TemporalFields;
...
+    type Output;
+    fn temporal_unwrap(self) -> TemporalResult<Self::Output>;
+}
+
+impl<T> TemporalUnwrap for Option<T> {
+    type Output = T;
+
+    fn temporal_unwrap(self) -> TemporalResult<Self::Output> {
+        self.ok_or(TemporalError::assert())
+    }
```
jedel1043, line 74: It would be good to add a `debug_assert`  here so that debug builds panic, and runtime builds throw the error.

jedel1043 (approved):
LGTM!

src/parsers.rs
```diff
@@ -0,0 +1,143 @@
...
+        );
+    }
+
+    Ok(record)
+}
+
+/// A utility function for parsing a `YearMonth` string
+pub(crate) fn parse_year_month(source: &str) -> TemporalResult<IxdtfParseRecord> {
+    let mut first_calendar: Option<Annotation> = None;
+    let mut critical_duplicate_calendar = false;
```
jedel1043, line 76: Is this required? I saw you assigned it but never used it after parsing.

jasonwilliams (approved):

PR #53, opened 2024-06-15
Update increment rounding functionality

nekevss:
This is still technically a draft. Although due to Duration rounding updates, I'd almost be in favor of merging this depending on feedback, primarily due to potential scope creep.

Overview:

`NormalizedTimeDuration` needs to be a `i128` and not a `f64` for safety as well as the new duration rounding algoritms. This means we need to be able to round both `i128` and `f64` for the library. So this branch currently updates the approach to rounding to allow handling of `i128` and `f64` at the same time, which allows `NormalizedTimeDuration` to be an `i128`.

@jedel1043 This is the rounding functionality I had mentioned on the GC call. If you have any thoughts on the API, let me know. There's probably an argument for a `PositiveIncrementRounder` that where `T` is a `u64` instead of adding `*_as_positive` methods on the current approach.

Tests / Scope Creep

~~Currently, this does break a couple unit tests for duration rounding, but I'm fairly certain that those have to do with truncation or some other issue in the duration rounding algorithm not the increment rounding itself (added some sanity unit tests to rounding), so I might look into this further.~~

~~This goes back to the scope creep I mentioned above. I'm trying not to make too much of an omnibus PR, and the duration rounding as a whole needs to be overhauled to the new algorithms. But the new algorithms need rounding and `NormalizedTimeDuration` updated, so I'm not entirely sure whether it's worth fixing the current tests with the current algorithm when the fixes can be completed while updating to the new algorithm. Not entirely sure, so any thoughts would be welcomed~~ :smile:

Actually found the bug. It was converting `NormalizedTimeDuration` -> `fractional_days`

### DIFF
--- a/src/components/duration/normalized.rs
+++ b/src/components/duration/normalized.rs
@@ -2,7 +2,11 @@
 
 use std::ops::Add;
 
-use crate::{options::TemporalRoundingMode, rounding::{IncrementRounder, Round}, TemporalError, TemporalResult, NS_PER_DAY};
+use crate::{
+    options::TemporalRoundingMode,
+    rounding::{IncrementRounder, Round},
+    TemporalError, TemporalResult, NS_PER_DAY,
+};
 
 use super::{DateDuration, TimeDuration};
 
@@ -51,11 +55,11 @@ impl NormalizedTimeDuration {
     pub(super) fn as_fractional_days(&self) -> f64 {
         // TODO: Verify Max norm is within a castable f64 range.
         let days: f64 = self.0.div_euclid(i128::from(NS_PER_DAY)) as f64;
-        let remainder:f64 = self.0.rem_euclid(i128::from(NS_PER_DAY)) as f64; 
+        let remainder: f64 = self.0.rem_euclid(i128::from(NS_PER_DAY)) as f64;
         days + (remainder / NS_PER_DAY as f64)
     }
 
-    // TODO: Potentially abstract sign into `Sign` 
+    // TODO: Potentially abstract sign into `Sign`
     /// Equivalent: 7.5.31 NormalizedTimeDurationSign ( d )
     #[inline]
     #[must_use]
@@ -82,7 +86,9 @@ impl NormalizedTimeDuration {
 
     /// Round the current `NormalizedTimeDuration`.
     pub(super) fn round(&self, increment: u64, mode: TemporalRoundingMode) -> TemporalResult<Self> {
-        let rounded = IncrementRounder::<i128>::from_potentially_negative_parts(self.0, increment.into()).round(mode);
+        let rounded =
+            IncrementRounder::<i128>::from_potentially_negative_parts(self.0, increment.into())
+                .round(mode);
         if rounded.abs() > MAX_TIME_DURATION {
             return Err(TemporalError::range()
                 .with_message("normalizedTimeDuration exceeds maxTimeDuration."));
--- a/src/components/duration/normalized.rs
+++ b/src/components/duration/normalized.rs
@@ -2,6 +2,8 @@
 
 use std::ops::Add;
 
+use num_traits::Euclid;
+
 use crate::{
     options::TemporalRoundingMode,
     rounding::{IncrementRounder, Round},
@@ -14,6 +16,7 @@ const MAX_TIME_DURATION: i128 = 9_007_199_254_740_991_999_999_999;
 
 // Nanoseconds constants
 
+const NS_PER_DAY_128BIT: i128 = NS_PER_DAY as i128;
 const NANOSECONDS_PER_MINUTE: f64 = 60.0 * 1e9;
 const NANOSECONDS_PER_HOUR: f64 = 60.0 * 60.0 * 1e9;
 
@@ -60,9 +63,8 @@ impl NormalizedTimeDuration {
     #[allow(unused)]
     pub(super) fn as_fractional_days(&self) -> f64 {
         // TODO: Verify Max norm is within a castable f64 range.
-        let days: f64 = self.0.div_euclid(i128::from(NS_PER_DAY)) as f64;
-        let remainder: f64 = self.0.rem_euclid(i128::from(NS_PER_DAY)) as f64;
-        days + (remainder / NS_PER_DAY as f64)
+        let (days, remainder) = self.0.div_rem_euclid(&NS_PER_DAY_128BIT);
+        days as f64 + (remainder as f64 / NS_PER_DAY as f64)
     }
 
     // TODO: Potentially abstract sign into `Sign`
--- a/src/components/duration/normalized.rs
+++ b/src/components/duration/normalized.rs
@@ -2,34 +2,36 @@
 
 use std::ops::Add;
 
-use crate::{options::TemporalRoundingMode, utils, TemporalError, TemporalResult, NS_PER_DAY};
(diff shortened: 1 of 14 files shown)


src/components/duration/normalized.rs
```diff
@@ -2,34 +2,46 @@
...
+        nanoseconds += (time.seconds * 1e9) as i128;
+        nanoseconds += (time.milliseconds * 1e6) as i128;
+        nanoseconds += (time.microseconds * 1_000.0) as i128;
+        nanoseconds += time.nanoseconds as i128;
         // NOTE(nekevss): Is it worth returning a `RangeError` below.
         debug_assert!(nanoseconds.abs() <= MAX_TIME_DURATION);
         Self(nanoseconds)
     }
 
+    // NOTE: `days: f64` should be an integer.
```
jasonwilliams, line 40: Is this going to be changed in a follow up? I guess there's too many call sites right now to change in this PR

src/components/duration/normalized.rs
```diff
@@ -39,42 +51,50 @@ impl NormalizedTimeDuration {
...
     #[inline]
     #[must_use]
     pub(super) fn sign(&self) -> i32 {
-        if self.0 < 0.0 {
-            return -1;
-        } else if self.0 > 0.0 {
-            return 1;
-        }
-        0
+        self.0.cmp(&0) as i32
```
jasonwilliams, line 73: Nice

src/rounding.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    fn result_ceil(&self) -> u64 {
+        // NOTE: Sanity debugs until proper unit tests to vet the below
+        debug_assert!(self.quotient_abs() < u64::MAX as i128);
+        self.quotient_abs() as u64 + 1
+    }
+
+    fn is_cardinal(&self) -> bool {
+        self.result_floor().rem_euclid(2) == 0
+    }
```
jasonwilliams, line 108: s this a correct name? Aren't all positive numbers cardinal rather than just even ones?

jasonwilliams (approved):
Im sure @jedel will know more about the changes, but LGTM

src/rounding.rs
```diff
@@ -0,0 +1,297 @@
...
+        }
+        rounded * self.divisor
+    }
+
+    fn round_as_positive(&self, mode: TemporalRoundingMode) -> u64 {
+        let unsigned_rounding_mode = mode.get_unsigned_round_mode(self.sign);
+        let rounded = apply_unsigned_rounding_mode(self, unsigned_rounding_mode);
+        rounded * self.divisor as u64
+    }
+}
```
jedel1043, line 79: I'd suggest making this an impl for `IncrementRounder<T: Roundable>` instead, and using any of the casting traits in [`num_traits`](https://docs.rs/num-traits/latest/num_traits/cast/index.html) to make it fully generic.

src/rounding.rs
```diff
@@ -0,0 +1,297 @@
...
+        self.dividend.rem_euclid(self.divisor) == 0
+    }
+
+    fn cmp_rem(&self) -> Option<Ordering> {
+        Some(
+            self.dividend
+                .rem_euclid(self.divisor)
+                .cmp(&self.divisor.div_euclid(2)),
+        )
+    }
```
jedel1043, line 92: Maybe a better name for the method? `cmp_rem` doesn't really say much about what the method needs to do, which is confusing since it just seems that it is comparing with half the divisor?

src/rounding.rs
```diff
@@ -0,0 +1,297 @@
...
+        // NOTE: Sanity debugs until proper unit tests to vet the below
+        debug_assert!(self.quotient_abs() < u64::MAX as i128);
+        self.quotient_abs() as u64
+    }
+
+    fn result_ceil(&self) -> u64 {
+        // NOTE: Sanity debugs until proper unit tests to vet the below
+        debug_assert!(self.quotient_abs() < u64::MAX as i128);
+        self.quotient_abs() as u64 + 1
+    }
```
jedel1043, line 104: Seems like these only differ from the f64 version by the assert? If this is the case, maybe the assert should be moved to `quotient_abs` instead, to remove these and simplify the API.

src/rounding.rs
nekevss: Yeah, the name doesn't match real well. The name was a bit more specification leaning, but definitely ambiguous.

src/components/duration/normalized.rs
nekevss: More than likely, yes. I'm pretty sure I wanted to be able to verify in all cases that it would be `days: u64` or some equivalent before making the change.

src/rounding.rs
nekevss: :confused: the only issue is that `Roundable` can't be implemented directly on `i128` it has to be implemented on the dividend/divisor pair

jedel1043: Yes, I'm just thinking that those operations don't really need an intermediate struct, they can just be methods that take the divisor by value.

`Roundable::is_exact(dividend: T, divisor: T) -> bool`

src/rounding.rs
```diff
@@ -0,0 +1,279 @@
...
+            assert!(unsigned_rounding_mode == TemporalUnsignedRoundingMode::HalfEven);
+            // 14. Let cardinality be (r1 / (r2 – r1)) modulo 2.
+            // 15. If cardinality is 0, return r1.
+            if Roundable::is_even_cardinal(dividend, divisor) {
+                return Roundable::result_floor(dividend, divisor);
+            }
+            // 16. Return r2.
+            Roundable::result_ceil(dividend, divisor)
+        }
+        None => unreachable!(),
```
jedel1043, line 204: Thought: Isn't this reachable if either of the operands is NaN?

src/rounding.rs
```diff
@@ -0,0 +1,279 @@
...
+
+pub(crate) trait Roundable:
+    Euclid + Div + PartialOrd + Signed + FromPrimitive + ToPrimitive + NumCast + ConstZero + Copy
+{
+    fn is_exact(dividend: &Self, divisor: &Self) -> bool;
+    fn compare_remainder(dividend: &Self, divisor: &Self) -> Option<Ordering>;
+    fn is_even_cardinal(dividend: &Self, divisor: &Self) -> bool;
+    fn result_floor(dividend: &Self, divisor: &Self) -> u64;
+    fn result_ceil(dividend: &Self, divisor: &Self) -> u64;
+    fn quotient_abs(dividend: &Self, divisor: &Self) -> Self {
```
jedel1043, line 21: All of those could take Self by value instead, since they require `Copy`. That would simplify the API a bit.

nekevss: Initially, I had those as value because of the copy, but wouldn't copying `i128`s be more expensive then passing the reference?

jedel1043: I'm pretty sure it's more costly to follow a pointer than to copy 8 additional bytes (additional indirection, the i128s won't be in the same cache line, etc.), but we can pick the simpler API and then benchmark in the future

nekevss: The change back isn't too difficult. I can change it to pass by value.

src/rounding.rs
jedel1043: Oh, that's nice. I'll rebase that other PR then.

jedel1043 (approved):
Overall it looks good. Maybe in the future we'll be able to move some invariants into the typesystem (some kind of `NonNanF64` would be great), but that can be added incrementally.

PR #54, opened 2024-06-16
Cleanup rounding increment usages with new struct

jedel1043:
Cleans up all usages of the rounding increment so that all have to pass through `RoundingIncrement`, increasing strictness.
Also moves the `RoundingIncrement` struct from `options.rs` to its own `increment.rs` file


nekevss (commented):
Can you move `RoundingIncrement` to a rounding.rs file. I'm hoping that may make the rebase a little easier. I'm also then not moving it over from `options.rs` to `rounding.rs`

nekevss (approved):

PR #62, opened 2024-06-25
Reject `IsoDate` when outside the allowed range

jedel1043:
Maybe we could move the check to `IsoDate`'s constructor? For now this is enough to pass test262.

### DIFF
--- a/src/components/date.rs
+++ b/src/components/date.rs
@@ -680,12 +680,6 @@ impl<C: CalendarProtocol> FromStr for Date<C> {
             ArithmeticOverflow::Reject,
         )?;
 
-        if !date.is_within_limits() {
-            return Err(
-                TemporalError::range().with_message("Date is not within ISO date time limits.")
-            );
-        }
-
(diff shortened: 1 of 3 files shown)


nekevss (approved):
I'd be fine with merging this as is, but I think I'd prefer this go on the constructor for `IsoDate` after taking a look at it.

The date from `IsoDate::new_unchecked` needs to be validated to be within limits on the constrain and reject options. Plus, that then allows implementing the check with `iso_dt_within_valid_limits`

jedel1043:
Yep, I'd also advocate for that, but we can do that on some future cleanups when we're certain that the spec won't suddenly change

nekevss:
Im not convinced that part of the spec will change. At least on the date and datetime constructor. Those are pretty stable.

jedel1043:
Hmm, in that case I'll just move this to the constructor.

nekevss (approved):
LGTM :)

PR #63, opened 2024-06-25
Ensure parsing throws with unknown critical annotations

jedel1043:


src/parsers.rs
```diff
@@ -38,13 +38,14 @@ fn parse_ixdtf(source: &str, variant: ParseVariant) -> TemporalResult<IxdtfParse
                 }
                 None => first_calendar = Some(annotation),
             }
+            return None;
         }
 
-        // Ignore all invalid annotations
-        None
+        // Make the parser handle any unknown annotation.
+        Some(annotation)
```
nekevss, line 45: Ah shoot! Nice catch.

nekevss (approved):

PR #65, opened 2024-06-30
Update duration rounding to new algorithms

### LINKED ISSUE #19: Implement `Duration` normalization and new algorithms

nekevss:
~~Currently a WIP.~~

This closes #19 outside of anything related to ZonedDateTime, which I'd prefer to consider separately due to the reliance on time zones.

This PR updates duration rounding to the new duration rounding algorithms.

As it's a bit of a larger PR, I thought I would create a draft in case anyone wants to take a look early.

Overall, this draft is fairly close to complete, but there are still at least a couple bugs around the actually rounding that I'm working through.


nekevss:
This PR should definitely wait until after #66 is merged so it can be rebased to remove the custom traits, but it is essentially ready for review outside of that impending rebase

There are a couple things to note:
  1. This implements the new rounding methods under a `round_v2` and doesn't remove the previous one just yet. It can be removed if everyone else thinks it should be in this PR, but I was sort of thinking about lagging on removing the previous methods while the rest is tested. Hopefully, it also makes the diff just a little bit easier to read. 
  2. Any rounding around `ZonedDateTime` is not included in this, but it's hard to implement and test that without proper time zone support, so if everyone else is in agreement I'd prefer to file that under a new issue.
  3. We should probably file an issue for more unit tests around the internal nudge algorithms and normalization records.

src/components/duration.rs
```diff
@@ -451,6 +452,225 @@ impl Duration {
         }
     }
 
+    #[inline]
+    pub fn round_v2<C: CalendarProtocol, Z: TzProtocol>(
+        &self,
+        increment: Option<RoundingIncrement>,
+        smallest_unit: Option<TemporalUnit>,
+        largest_unit: Option<TemporalUnit>,
+        rounding_mode: Option<TemporalRoundingMode>,
```
jedel1043, line 461: Seeing how many optionals are here, maybe we should extract this into a `RoundingOptions` struct that implements `Default`.

src/components/duration.rs
```diff
@@ -451,6 +452,225 @@ impl Duration {
         }
     }
 
+    #[inline]
+    pub fn round_v2<C: CalendarProtocol, Z: TzProtocol>(
```
jedel1043, line 456: About the `round_v2`, I think it should just replace the old rounding function. We just need to test `Boa` against the PR branch to see that everything looks good.

src/components/duration/normalized.rs
```diff
@@ -118,21 +145,636 @@ impl Add<Self> for NormalizedTimeDuration {
...
+    }
+
+    pub(crate) fn norm(&self) -> NormalizedTimeDuration {
+        self.0 .1
+    }
+
+    pub(crate) fn sign(&self) -> TemporalResult<i32> {
+        if self.0 .0.sign() != self.0 .1.sign() {
+            return Err(TemporalError::range().with_message("Invalid NormalizedDuration signs."));
+        }
```
jedel1043, line 185: Can we move this to the constructor?

src/components/duration/normalized.rs
```diff
@@ -118,21 +145,636 @@ impl Add<Self> for NormalizedTimeDuration {
...
-/// A normalized `DurationRecord` that contains a `DateDuration` and `NormalizedTimeDuration`.
+// ==== NormalizedDurationRecord ====
+//
+// A record consisting of a DateDuration and NormalizedTimeDuration
+//
+
+/// A NormalizedDurationRecord is a duration record that contains
+/// a `DateDuration` and `NormalizedTimeDuration`.
 #[derive(Debug, Clone, Copy)]
 pub struct NormalizedDurationRecord(pub(crate) (DateDuration, NormalizedTimeDuration));
```
jedel1043, line 156: I think it would be better to use a proper struct instead of a tuple here. Also, can we make this private? I'm not sure if we need to allow external mutations here, seeing that we require checks on `new` that may not be preserved by external code.

src/utils.rs
```diff
@@ -1,9 +1,27 @@
...
+        if matches!(ordering, Ordering::Less) {
+            return Self(ordering);
+        }
+        Self(Ordering::Greater)
+    }
+
+    pub(crate) fn get(&self) -> i8 {
+        self.0 as i8
+    }
+}
```
jedel1043, line 23: Seeing that this is only used by `sign`, maybe it can just be a wrapper around `NonZeroI8`?
```rust
struct Sign(NonZeroI8);

impl Sign {
    const POSITIVE: Self = const {
        let Some(n) = NonZeroI8::new(1) else {
            panic!("invalid number for sign");
        }
        Self(n)
    }

    const NEGATIVE: Self = const {
        let Some(n) = NonZeroI8::new(-1) else {
            panic!("invalid number for sign");
        }
        Self(n)
    }

    pub(crate) fn get(&self) -> i8 {
        self.0.get()
    }
}
```
This also makes `std::mem::size_of::<Option<Sign>>() == std::mem::size_of::<i8>()`

jedel1043 (commented):
Looks great! I have some suggestions on some APIs that we can improve.

src/components/duration.rs
nekevss: Yeah, I was thinking about doing this in a follow up, but it could be added to this as well.

Temporal spec refers to them as difference settings `13.53 GetDifferenceSettings`, but I don't hate `RoundingOptions` as a name. Any preference between `DifferenceSettings` vs. `RoundingOptions`

jedel1043: I prefer `RoundingOptions` for consistency. If we need `DifferenceOptions` in some other place (the diff_instant operation maybe?), we can do a `type DifferenceOptions = RoundingOptions;` for it.

src/utils.rs
jedel1043: Actually, an `enum Sign` optimizes perfectly fine, apparently.
```rust
enum Sign {
    Positive = 1,
    Negative = -1,
}

dbg!(std::mem::size_of::<Sign>()); // 1
dbg!(std::mem::size_of::<Option<Sign>>()); // 1
```

nekevss: Hmmm maybe a Sign, which can be zero, and a NonZeroSign (hence the NonZeroOrdering).

jedel1043: Do we need a sign that can be zero though? I would expect it to only have Positive and Negative, since the zero case is not really useful IIRC.

nekevss: Temporal Spec has a -1, 0, 1 return

https://tc39.es/proposal-temporal/#sec-temporal-durationsign

src/components/duration.rs
nekevss: From what I can tell they both essentially resolve to the same options but with different defaults and methodology. Updated them to a public `DifferenceSettings` and `RoundingOptions` that resolve to `ResolvedRoundingOptions`, let me know what you think!

src/components/duration.rs
```diff
@@ -447,68 +430,37 @@ impl Duration {
...
 
-    /// Rounds the current `Duration`.
     #[inline]
     pub fn round(
         &self,
-        increment: Option<RoundingIncrement>,
-        smallest_unit: Option<TemporalUnit>,
-        largest_unit: Option<TemporalUnit>,
-        rounding_mode: Option<TemporalRoundingMode>,
+        options: Option<RoundingOptions>,
```
jedel1043, line 436: I would make this take the `RoundingOptions` directly, because in the case where someone wants to set an option, they would need to double-wrap with `Some`, which is not ideal.

src/options.rs
```diff
@@ -4,15 +4,166 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct ResolvedRoundingOptions {
+    pub(crate) largest_unit: TemporalUnit,
+    pub(crate) smallest_unit: TemporalUnit,
+    pub(crate) increment: RoundingIncrement,
+    pub(crate) rounding_mode: TemporalRoundingMode,
+}
```
jedel1043, line 60: Praise: Loving this new API! I think it really makes sense to have an options struct that resolves all the `Option`s in `RoundingOptions`, instead of having to `unwrap` everywhere.

src/components/duration.rs
```diff
@@ -22,6 +23,40 @@ pub use date::DateDuration;
...
+
+impl From<Ordering> for DurationSign {
+    fn from(value: Ordering) -> Self {
+        match value {
+            Ordering::Greater => Self::Positive,
+            Ordering::Equal => Self::Zero,
+            Ordering::Less => Self::Negative,
+        }
+    }
+}
```
jedel1043, line 42: Hmmm, I don't know if it makes much sense to have a conversion from these enums. `Ordering` is a completely different concept than `DurationSign`, and things like `DurationSign::from(zero.cmp(&target))` don't make much sense if you think about it.

src/components/duration/time.rs
```diff
@@ -365,6 +370,69 @@ impl TimeDuration {
 // ==== TimeDuration method impls ====
 
 impl TimeDuration {
+    // TODO: Maybe move to `NormalizedTimeDuration`
+    pub fn round_v2(
```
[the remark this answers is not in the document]
nekevss: I went to remove it, but there's still a dependency on the original round. I was going to follow up on this in a different PR.

src/components/duration.rs
nekevss: Yeah, it's a bit awkward. I wasn't super in love with it. Maybe just a `From<i8>`

src/components/duration.rs
```diff
@@ -22,6 +23,40 @@ pub use date::DateDuration;
...
+
+impl From<DurationSign> for NonZeroDurationSign {
+    fn from(value: DurationSign) -> Self {
+        if matches!(value, DurationSign::Negative) {
+            return Self::Negative;
+        }
+        Self::Positive
+    }
+}
+
```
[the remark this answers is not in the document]
nekevss: I'm going back and forth with `DurationSign` vs. `Sign`. :confused: Updated to `Sign` this time around.

src/options.rs
```diff
@@ -4,15 +4,166 @@
...
+
+impl ResolvedRoundingOptions {
+    pub(crate) fn from_diff_settings(
+        options: DifferenceSettings,
+        operation: DifferenceOperation,
+        fallback_largest: TemporalUnit,
+        fallback_smallest: TemporalUnit,
+    ) -> TemporalResult<(f64, Self)> {
```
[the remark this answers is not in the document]
nekevss: Updated to a `Sign` value.

src/components/duration/normalized.rs
```diff
@@ -118,20 +144,607 @@ impl Add<Self> for NormalizedTimeDuration {
...
+        Self::new(date, NormalizedTimeDuration::default())
+    }
+
+    pub(crate) fn date(&self) -> DateDuration {
+        self.date
+    }
+
+    pub(crate) fn norm(&self) -> NormalizedTimeDuration {
+        self.norm
+    }
```
jedel1043, line 182: Same suggestion of using the full name.

src/lib.rs
```diff
@@ -80,6 +82,34 @@ impl<T> TemporalUnwrap for Option<T> {
...
+}
+
+impl Sign {
+    /// Coerces the current `Sign` to be either negative or positive.
+    pub(crate) fn as_non_zero(&self) -> Self {
+        if matches!(self, Self::Zero) {
+            return Self::Positive;
+        }
+        *self
+    }
```
jedel1043, line 110: If this is to multiply a duration with the sign, what about having an `as_sign_multiplier` function instead that returns `1i8` for zero and positive, and `-1i8` for negative?

nekevss: It is. It's specifically a nonzero variant in these rounding methods vs. the regular multiplication that occurs elsewhere. I'm fine with `as_sign_multiplier` though. Better than anything I'm coming up with in this instance.

jedel1043 (approved):
Really impressive work! Pretty much perfect IMO :)

jasonwilliams (approved):
LGTM!
As @jedel said this looks great, big improvement

PR #66, opened 2024-07-01
Remove `CalendarProtocol` and `TimeZoneProtocol`

### LINKED ISSUE #64: Remove custom calendars & timezones
After discussion in the last TC39 plenary, it was confirmed there won't be any custom calendar usage with Temporal. Even if this was to return in a V2, it wouldn't come back in the same shape we see today. Functionality could be injected at a lower level (similar to Timezones with tzdatabase).

Despite this being a library to power higher-level ECMAScript implementations it should still try to follow the Temporal specification as much as possible and not include things the spec doesn't make use of. We can also add extra functionality further down the line in other areas.

jedel1043:
Closes #64.

Change due thanks to the spec being simplified by removing those APIs. We could preserve them, but it would be better to follow the spec first, then add extensions in the future.

This cleans up a significant amount of code, and also unifies the context-free and context-dependant APIs into a single one; we don't require additional context now.

This also simplifies our calendar API by storing pointers to the constructed calendars instead.


src/components/calendar.rs
```diff
@@ -52,6 +65,154 @@ pub const CALENDAR_PROTOCOL_METHODS: [&str; 21] = [
     "yearOfWeek",
 ];
 
+#[derive(Debug, Clone)]
+pub struct TemporalCalendar(Ref<'static, AnyCalendar>);
```
nekevss, line 69: thought: Should this be `Calendar` or `TemporalCalendar`

Potential bikeshed. We don't really refer to any other component as `Temporal*` outside of `TemporalRoundingMode`. My thought is that at minimum the names should be consistent. `TemporalCalendar`, `TemporalDate`, etc. or `Calendar`, `Date`, etc.

While I can see the argument for a `Temporal` prefix on the options as they may be present across libraries, I'm not sure the potential for colliding names being present on the components.

nekevss (commented):
This is going to simplify the API so much! A little bittersweet but super happy to see the API simplified and the context APIs removed.

Had a couple comments, nothing too much.

nekevss (approved):
LGTM

jasonwilliams (approved):
This simplifies the whole library, great changes.
LGTM

PR #70, opened 2024-07-02
WIP: Build out support for `Now` component

nekevss:
Building out the `Now` component.

This is going to be minimal support for the component.

There's also a question around how we should handle fetching system calls in `temporal_rs` (although this PR starts by using an external crate)


nekevss:
Closing in favor of #99 & #100

PR #73, opened 2024-07-06
Update duration validation and switch asserts to debug-asserts

nekevss:
This updates the `is_valid_duration` check to the current specification and also changes a couple `asserts` into `debug_asserts` to prevent panics.


jedel1043 (commented):
Should we add a new `temporal_assert`, similar to `temporal_unwrap` that throws an error on release?

nekevss:
Yeah, that's not a bad approach. If we added an assert with a log, it might make debugging easier too. I'll add it to this quick.

jedel1043 (approved):
LGTM!

raskad (approved):

PR #74, opened 2024-07-07
Implement add and subtract methods for Duration

nekevss:
Related to ongoing work for #11

As it says in the title, implementing the add and subtract methods for duration.

### DIFF
--- a/src/components/duration.rs
+++ b/src/components/duration.rs
@@ -100,6 +100,40 @@ impl Duration {
             .map(|x| TemporalUnit::from(10 - x.0))
             .unwrap_or(TemporalUnit::Nanosecond)
     }
+
+    /// The internal implementation of adding/subtracting `Duration`s
+    #[inline]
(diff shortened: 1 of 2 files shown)


src/components/duration.rs
```diff
@@ -400,6 +434,18 @@ impl Duration {
         }
     }
 
+    /// Returns the result of adding a `Duration` to the current `Duration`
+    #[inline]
+    pub fn add(&self, other: &Self) -> TemporalResult<Self> {
+        self.internal_add(other)
```
jedel1043, line 440: Nit: Dldo we need to have an `internal_add` method? I think it would be cleaner to move the code of `internal_add` to `add`, then just call `self.add(other.negated())` for `subtract`.

jedel1043 (commented):
Everything looks great! I just have a small nit.

jedel1043 (approved):

PR #75, opened 2024-07-07
Implement PartialEq and Eq for `Calendar`, `Date`, and `DateTime`

nekevss:
Per title, this implements `PartialEq` and `Eq` for `Calendar` and some corresponding components. This should allow us to bypass the `equals` method.

There is a question for API ergonomics whether we provide an explicit `equals` method, but I'm not entirely sure that's needed with the trait implementations and feels counterproductive to a Rust API.

### DIFF
--- a/src/components/calendar.rs
+++ b/src/components/calendar.rs
@@ -69,6 +69,14 @@ impl Default for Calendar {
     }
 }
 
+impl PartialEq for Calendar {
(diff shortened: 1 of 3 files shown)


jedel1043 (approved):
Looks good! Agreed that `equals` should be a separate change to discuss, but implementing `Eq` is a lot less controversial.

PR #76, opened 2024-07-07
Bump `temporal_rs` version and update changelog

nekevss:
This PR bumps the version in the changelog and bumps the `Cargo.toml` version to `0.0.3` for release


jedel1043:
Was thinking that we should probably exclude dependabot updates from git cliff

nekevss:
Potentially? it does look like we may be able to group the `CHANGELOG.md` better if we used conventional commits.

CHANGELOG.md
```diff
@@ -22,14 +18,9 @@
 * Patch `(un)balance_relative` to avoid panicking by @jedel1043 in [#48](https://github.com/boa-dev/temporal/pull/48)
 * Cleanup rounding increment usages with new struct by @jedel1043 in [#54](https://github.com/boa-dev/temporal/pull/54)
 * Add struct to encapsulate invariants of rounding increments by @jedel1043 in [#49](https://github.com/boa-dev/temporal/pull/49)
-* Bump icu_calendar from 1.5.0 to 1.5.1 by @dependabot[bot] in [#52](https://github.com/boa-dev/temporal/pull/52)
-* Bump tinystr from 0.7.5 to 0.7.6 by @dependabot[bot] in [#51](https://github.com/boa-dev/temporal/pull/51)
 * Migrate parsing to `ixdtf` crate by @nekevss in [#50](https://github.com/boa-dev/temporal/pull/50)
-* Bump `icu_calendar` to 1.5 by @jedel1043 in [#47](https://github.com/boa-dev/temporal/pull/47)
```
nekevss: For example, this was excluded under the "^Bump" rule, which may not be the best approach.

jedel1043:
> EDIT: So currently it now filters out any commit that is "^Bump", but I think we could configure "dependabot"'s commit message to use conventional commits, something like "chore(dependency): ",which may be better overall.

Makes sense. We can add that on a separate PR I think.

jedel1043 (approved):
Changelog looks good to me!

PR #77, opened 2024-07-08
Add commit-message prefix to dependabot

nekevss:
As mentioned in #76, this should hopefully allow us to filter bot commits from future changelogs.

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -69,7 +69,6 @@ commit_preprocessors = [
 ]
 # Commit grouping 
 commit_parsers = [
-  { message = "^Bump", skip = true },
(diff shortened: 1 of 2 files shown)


cliff.toml
```diff
@@ -71,6 +71,7 @@ commit_preprocessors = [
 commit_parsers = [
   { message = "^Bump", skip = true },
```
jedel1043, line 72: Remove this?

nekevss: I was keeping it in just in case, but it can be removed.

jedel1043 (approved):

PR #81, opened 2024-07-12
Add some missing trait implementations

nekevss:
Mostly adding some missing `From` implementations according to values that may be convertible, and a `FromStr` for `Time`.


jedel1043 (approved):
Looks great!

PR #82, opened 2024-07-16
Implement compare functionality and some more traits

nekevss:
This adds some `compare` functions to `Date` and `DateTime`, and Ordering for `ZonedDateTime`.

The compare functions are notably different from `Eq` and `PartialEq` as they compare the internal ISO representation ignoring the Calendar


jedel1043 (approved):
Looks good!

PR #83, opened 2024-07-16
Implement `DateTime` diffing methods `Until` and `Since`

nekevss:
This implements diffing for `DateTime` and the corresponding `since` and `until` methods.

TODO: add basic unit tests for the methods.

### DIFF
--- a/src/components/datetime.rs
+++ b/src/components/datetime.rs
@@ -3,7 +3,10 @@
 use crate::{
     components::{calendar::Calendar, duration::TimeDuration, Instant},
     iso::{IsoDate, IsoDateSlots, IsoDateTime, IsoTime},
-    options::{ArithmeticOverflow, ResolvedRoundingOptions, TemporalUnit},
+    options::{
+        ArithmeticOverflow, DifferenceOperation, DifferenceSettings, ResolvedRoundingOptions,
+        TemporalUnit,
+    },
     parsers::parse_date_time,
     temporal_assert, TemporalError, TemporalResult, TemporalUnwrap,
 };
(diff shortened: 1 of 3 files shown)


src/components/datetime.rs
```diff
@@ -89,6 +92,43 @@ impl DateTime {
...
+        if self.iso == other.iso {
+            return Ok(Duration::default());
+        }
+
+        // Step 10-11.
+        let (result, _) = self.diff_dt_with_rounding(other, options)?;
+
+        // Step 12
+        match sign {
+            crate::Sign::Positive => Ok(result),
```
jedel1043, line 126: You can avoid the panic by also matching against `Sign::Zero` here.

src/components/datetime.rs
```diff
@@ -543,4 +598,61 @@ mod tests {
...
+
+    fn create_diff_setting(
+        smallest: TemporalUnit,
+        increment: u32,
+        rounding_mode: TemporalRoundingMode,
+    ) -> DifferenceSettings {
+        DifferenceSettings {
+            largest_unit: None,
+            smallest_unit: Some(smallest),
+            increment: Some(unsafe { RoundingIncrement::new_unchecked(increment) }),
```
jedel1043, line 610: I would prefer if we didn't use `unsafe` here. We can just use `try_new` and unwrap.

jedel1043 (changes requested):
Nice work! We should avoid using `unsafe` where it's not required.

jedel1043 (approved):
Nice job!

PR #84, opened 2024-07-17
Add `with_*` methods to `Date` and `DateTime`

nekevss:
Per title, this PR adds `with_calendar` and `with_time` methods to `Date` and `DateTime`. This also adds some `From` impls for `Calendar`.

### DIFF
--- a/src/components/date.rs
+++ b/src/components/date.rs
@@ -193,13 +193,18 @@ impl Date {
         Ok(Self::new_unchecked(iso, calendar))
     }
 
-    #[must_use]
-    /// Creates a `Date` from a `DateTime`.
-    pub fn from_datetime(dt: &DateTime) -> Self {
-        Self {
-            iso: dt.iso_date(),
-            calendar: dt.calendar().clone(),
-        }
+    /// Creates a
+    pub fn with_calendar<I>(&self, calendar_like: Option<I>) -> TemporalResult<Self>
+    where
+        I: Into<Calendar>,
+    {
+        Self::new(
+            self.iso_year(),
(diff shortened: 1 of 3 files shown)


src/components/date.rs
```diff
@@ -193,13 +193,18 @@ impl Date {
...
 
-    #[must_use]
-    /// Creates a `Date` from a `DateTime`.
-    pub fn from_datetime(dt: &DateTime) -> Self {
-        Self {
-            iso: dt.iso_date(),
-            calendar: dt.calendar().clone(),
-        }
+    /// Creates a
+    pub fn with_calendar<I>(&self, calendar_like: Option<I>) -> TemporalResult<Self>
```
jedel1043, line 197: I don't think we should accept an `Option` for something that only requires a calendar. We should just require users to pass the calendar directly.

nekevss: That's fair. The intent in this instance was to try and mimic the proposal as much as possible around accepting a `calendarLike` vs. a `calendar`. But there's not really any negative to restricting to just a calendar. It's probably a better API too.

jedel1043: I think the spec API is fine, but in my opinion the APIs should be adapted to the explicitness of Rust APIs instead of having a 1-1 equivalent in both, which in this case means having some function that people can call to get the calendar before calling the function, instead of filling the API with generics.

jedel1043: This is partly because any APIs that have generics in them will make it really difficult to implement FFI interfaces for them.

jedel1043 (approved):
Looks great!

PR #85, opened 2024-07-19
Updates to instant and its methods

nekevss:
This PR updates Instant to some of the newer specification changes. 

TLDR of the below:
- Updates `Instant` struct
- Updates diffing ops and adds tests
- Implements `FromStr`.

Primarily, it brings `since` and `until` up to date with both `temporal_rs` and the specification as a whole (along with at least 2 basic tests for `since` and `until`)

This PR also changes Instant inner nanoseconds from `BigInt` to `i128`. My thought is to eventually remove all usage of `BigInt` from `temporal_rs` in general. Primarily motivated by the thought that if `NormalizedTimeDuration` is using a `i128`, I don't see a huge reason why `Instant` should differ, unless I'm overlooking something and someone thinks otherwise.

Finally, it implements `FromStr` for `Instant`. This will allow parsing of `JsString` in Boa, and would unblock `to_temporal_instant`.


src/components/instant.rs
```diff
@@ -278,35 +270,240 @@ impl Instant {
...
+impl FromStr for Instant {
+    type Err = TemporalError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let ixdtf_record = parse_instant(s)?;
+
+        // NOTE: parse_instant rejects on missing required fields. So the below
+        // `temporal_unwraps` MUST be present.
+
+        // Find the IsoDate
+        let date = ixdtf_record.date.temporal_unwrap()?;
```
jedel1043, line 286: Hmm, is this asserting an invariant? Because `temporal_unwrap` should only be used to ensure invariants are preserved. If this is used to unwrap error caused by the user, then I think it should return an error instead, which is a bit more explanatory.

nekevss: It should be an invariant at this point. `parse_instant` requires a string with a `DateRecord`, `TimeRecord`, and `OffsetRecord`, and throws an error if any of them are `None`.

We can return an error in these places again, but they would be redundant errors that realistically would never throw.

jedel1043: Oh, now I see why. On `parse_instant`, we validate that the fields are `Some` and throw otherwise. Can we move that invariant to the typesystem? Instead of returning an `IxdtfRecord`, we return `IxdtfDurationRecord` with the validated fields being not `Option`.

jedel1043 (approved):
Looks great!

PR #86, opened 2024-07-19
Update `Duration` types to use a `FiniteF64` instead of `f64` primitive.

nekevss:
This PR updates `Duration` types to have `FiniteF64` fields instead of just a `f64`.

I still have to re-implement a `PartialDurationRecords` struct but thought I'd post this as a draft in the meantime for feedback/thoughts.

Also, `FiniteF64` probably shouldn't live in `utils`, but I wasn't sure exactly where to move it... maybe a primitives mod?


src/components/duration/time.rs
```diff
@@ -198,19 +200,20 @@ impl TimeDuration {
         // 12. Return ! CreateTimeDurationRecord(days × sign, hours × sign, minutes × sign, seconds × sign, milliseconds × sign, microseconds × sign, nanoseconds × sign).
         let days = (days as i64).mul_add(sign.into(), 0);
         let result = Self::new_unchecked(
-            (hours as i32).mul_add(sign, 0).into(),
-            (minutes as i32).mul_add(sign, 0).into(),
-            (seconds as i32).mul_add(sign, 0).into(),
-            (milliseconds as i32).mul_add(sign, 0).into(),
-            (microseconds as i32).mul_add(sign, 0).into(),
-            (nanoseconds as i32).mul_add(sign, 0).into(),
+            FiniteF64::from((hours as i32).mul_add(sign, 0)),
```
jedel1043, line 203: This sounds like a perfect fit for [copysign](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.copysign).

src/components/duration.rs
```diff
@@ -825,16 +795,16 @@ impl FromStr for Duration {
...
-            f64::from(months) * sign,
-            f64::from(weeks) * sign,
-            f64::from(days) * sign,
-            hours * sign,
-            minutes * sign,
-            seconds * sign,
-            millis * sign,
-            micros * sign,
-            nanos * sign,
+            FiniteF64::from(years).unchecked_mul(sign),
```
jedel1043, line 798: Same suggestion about `copysign`.

src/utils.rs
```diff
@@ -1,9 +1,151 @@
...
 
+#[derive(Debug, Default, Clone, Copy, PartialEq, PartialOrd)]
+pub struct FiniteF64(pub(crate) f64);
+
+impl FiniteF64 {
+    #[inline]
+    #[must_use]
+    pub unsafe fn new_unchecked(value: f64) -> Self {
+        Self(value)
+    }
```
jedel1043, line 17: Does this need to be unsafe? We're not applying any memory optimizations that depend on f64 being finite, right?

jedel1043 (commented):
Really liking the new type!

src/utils.rs
nekevss: Not necessarily for memory optimizations, but the idea of it marking it as unsafe was primarily because it's an unchecked construction. For instance, someone might do something like:

```rust
let totally_finite = FiniteF64::new_unchecked(f64::NAN);
```

It totally undercuts the invariance of the type, so to make the unsafe is meant more to mark that unchecked API public. That being said, I'd be fine with just getting rid of the method entirely. As I'm not entirely sold on having the method in general (I had started with it there, and as I worked on this, the less I thought it was needed).

jedel1043 (approved):
LGTM! Just needs a rebase.

PR #87, opened 2024-07-20
Refactor `TemporalFields` interface and add `FieldsKey` enum

nekevss:
This refactors the representation to `TemporalFields`. The previous implementation was a bit cumbersome. This is meant to trim down the implementation and, hopefully, make the API a simpler.

### DIFF
--- a/src/fields.rs
+++ b/src/fields.rs
@@ -237,6 +237,12 @@ impl TemporalFields {
         }
     }
 
+    /// Sets a field as active. This will require the field's default value to be used if the field is not yet set.
+    #[inline]
+    pub fn activate_field(&mut self, key: FieldKey) {
+        self.bit_map.set(key.into(), true);
+    }
+
     /// Gets the value of a `FieldKey` if the field has been set to active. If the field
     /// has not been set, then return `None`.
     #[inline]
--- a/src/fields.rs
+++ b/src/fields.rs
@@ -1,25 +1,83 @@
 //! This module implements a native Rust `TemporalField` and components.
 
-use std::{collections::hash_map::Keys, str::FromStr};
+use core::fmt;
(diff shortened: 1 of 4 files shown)


src/fields.rs
```diff
@@ -113,516 +138,168 @@ impl FromStr for FieldConversion {
...
-            millisecond: 0,
-            microsecond: 0,
-            nanosecond: 0,
-            offset: None,
-            era: None,
-            era_year: None,
-            time_zone: None,
-        }
-    }
+    properties: FxHashMap<FieldKey, FieldValue>,
```
jedel1043, line 143: Overall question. Why do we need to use a hashmap? I would say having a map-like interface for the current struct approach would also work well, with the advantage of being all in the stack.

nekevss: Hmmmm, changing the interface might work as well.

I'm not entirely sure how much of a fan I am of the large struct that this switches away from, but it could also primarily be the API for the struct that I'm annoyed by.

I'll play around with it a bit more :smile:

nekevss:
Marking as ready for review. I think this is an improvement interface/API wise vs. the old methods. If anyone has suggestions, let me know.

jedel1043 (commented):
Maybe this sounds paradoxical after saying in the last PR that we should reduce our usage of generics, but what do you think about using generics to move the errors to user code?

It would be cool to have something like
```rust
fields.set::<Month>(5i32);
```

nekevss:
Hmmmm, I do sort like the idea. If it makes the API easy to use and the code a tad cleaner / easier to reason through, then I don't see a reason to not try.

jedel1043 (approved):
Nice! I think we can merge this as it is, then discuss the new design on Matrix and create a new PR for the redesign

PR #88, opened 2024-07-21
Implement `DateTime` round method

nekevss:
This PR implements `DateTime:round`. In order to do this, `IsoDateTime::round` needed to be implemented, and `IsoTime::round` was updated to the current specification version

### DIFF
--- a/src/iso.rs
+++ b/src/iso.rs
@@ -175,7 +175,11 @@ impl IsoDateTime {
 
     pub(crate) fn round(&self, resolved_options: ResolvedRoundingOptions) -> TemporalResult<Self> {
         let (rounded_days, rounded_time) = self.time.round(resolved_options)?;
-        let balance_result = IsoDate::balance(self.date.year, self.date.month.into(), i32::from(self.date.day) + rounded_days);
+        let balance_result = IsoDate::balance(
+            self.date.year,
+            self.date.month.into(),
+            i32::from(self.date.day) + rounded_days,
+        );
         Self::new(balance_result, rounded_time)
     }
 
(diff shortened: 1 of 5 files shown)


src/iso.rs
```diff
@@ -654,140 +660,146 @@ impl IsoTime {
...
+                i128::from(self.nanosecond)
             }
-            // 8. Else,
-            // a. Assert: unit is "nanosecond".
-            // b. Let quantity be nanosecond.
-            TemporalUnit::Nanosecond => u64::from(self.nanosecond),
             _ => {
                 return Err(TemporalError::range()
-                    .with_message("Invalid temporal unit provided to Time.round."))
+                    .with_message("Invalid smallestUNit value for time rounding."))
```
jedel1043, line 716:
```suggestion
                    .with_message("Invalid smallestUnit value for time rounding."))
```

jedel1043 (approved):

src/iso.rs
```diff
@@ -665,140 +675,146 @@ impl IsoTime {
...
+                // a. Let quantity be ((((hour × 60 + minute) × 60 + second) × 1000 + millisecond)
+                // × 1000 + microsecond) × 1000 + nanosecond.
+                ((((i128::from(self.hour) * 60 + i128::from(self.minute)) * 60
+                    + i128::from(self.second))
+                    * 1000
+                    + i128::from(self.millisecond))
+                    * 1000
+                    + i128::from(self.microsecond))
+                    * 1000
+                    + i128::from(self.nanosecond)
```
jedel1043, line 692: I think this could benefit from separating it on several variables, even if they're just used once. Also applies to the other branches that do a big operation like this.

nekevss: Yeah, it would probably help with the readability.

jedel1043 (approved):
Nice work!

PR #89, opened 2024-07-23
Implement `MonthCode`, `PartialDate`, and `Date::with`

nekevss:
So the title basically says it all as far as implementation. That being said, there's a lot of interpretation going into this that I'd like feedback on if possible, and it could be the right direction or it could be the wrong direction (although I'm leaning the former over the latter).

The main culprit is basically [PrepareTemporalFields](https://tc39.es/proposal-temporal/#sec-temporal-preparetemporalfields).

Background for discussion/sanity check:

Up until recently, I basically thought it would be an implementation detail on the engine/interpreter side, but the thing that always bugged me was the `requiredFields` parameter, being either a List or PARTIAL. We could probably do that to specification, but we might be providing something like `Some(Vec::default())` as an argument, and it basically just felt clunky.

After the recent `TemporalFields` update, I went to implement the `TemporalFields` portion of the `toX` abstract ops in Boa and realized that PARTIAL is never called in the `toX` operations, and it's actually exclusively called in `with` methods. We already have a sort of precedence for partials with `PartialDuration`.

There's some benefits to this: we can have a with method on the native rust side, ideally the complexity that exists in `PrepareTemporalFields` can be made a bit easier to reason about.

Potential negatives: we might end up deviating from the specification as far as the order of when errors are thrown and observability (TBD...potentially a total non-issue) and this is probably opening up a can of worms around what would be the ideal API for a `PartialDate`, `PartialDateTime`, and `PartialTime`.

That all being said, I think the benefits do most likely outweigh any negatives, and it would be really cool to have `with` method implementations. I'm just not entirely sure around the API.

Also, there's an addition of a `MonthCode` enum to make `From<X> for TemporalFields` implementations easier.

### DIFF
--- a/src/components/calendar.rs
+++ b/src/components/calendar.rs
@@ -638,7 +638,7 @@ impl Calendar {
         &self,
         keys: &[TemporalFieldKey],
     ) -> TemporalResult<Vec<TemporalFieldKey>> {
-        let mut ignored_keys = Vec::default();
+        let mut ignored_keys = Vec::with_capacity(5);
         if self.is_iso() {
             // NOTE: It is okay for ignored keys to have duplicates?
             for key in keys {
--- a/src/components/calendar.rs
+++ b/src/components/calendar.rs
@@ -10,7 +10,7 @@ use crate::{
         duration::{DateDuration, TimeDuration},
         Date, DateTime, Duration, MonthDay, YearMonth,
     },
-    fields::{TemporalFieldKey, TemporalFields},
+    fields::{FieldMap, TemporalFields},
     iso::{IsoDate, IsoDateSlots},
     options::{ArithmeticOverflow, TemporalUnit},
     TemporalError, TemporalResult,
@@ -634,19 +634,16 @@ impl Calendar {
     }
 
     /// Provides field keys to be ignored depending on the calendar.
-    pub fn field_keys_to_ignore(
-        &self,
-        keys: &[TemporalFieldKey],
-    ) -> TemporalResult<Vec<TemporalFieldKey>> {
-        let mut ignored_keys = Vec::with_capacity(5);
+    pub fn field_keys_to_ignore(&self, keys: FieldMap) -> TemporalResult<FieldMap> {
+        let mut ignored_keys = FieldMap::empty();
         if self.is_iso() {
             // NOTE: It is okay for ignored keys to have duplicates?
-            for key in keys {
-                ignored_keys.push(*key);
-                if key == &TemporalFieldKey::Month {
-                    ignored_keys.push(TemporalFieldKey::MonthCode)
-                } else if key == &TemporalFieldKey::MonthCode {
-                    ignored_keys.push(TemporalFieldKey::Month)
+            for key in keys.iter() {
+                ignored_keys.set(key, true);
+                if key == FieldMap::MONTH {
+                    ignored_keys.set(FieldMap::MONTH_CODE, true);
+                } else if key == FieldMap::MONTH_CODE {
+                    ignored_keys.set(FieldMap::MONTH, true);
                 }
             }
 
@@ -700,8 +697,9 @@ impl From<YearMonth> for Calendar {
 
 #[cfg(test)]
 mod tests {
+    use crate::{components::Date, iso::IsoDate, options::TemporalUnit};
 
-    use super::*;
+    use super::Calendar;
 
     #[test]
     fn date_until_largest_year() {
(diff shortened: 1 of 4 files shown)


ptomato:
PrepareTemporalFields has always been slightly weird. I am trying to see if it can be simplified if custom calendars are not a thing.

nekevss:
I think it's definitely can be split apart. Does the split between a "PartialDateLike" (so to speak) in `with` methods vs. a required "DateLike" JsObject in the `to_temporal_x` abstract make sense to the general intent of the proposal? Or am I overlooking something from those codepaths.

ptomato:
Those two operations were combined into PrepareTemporalFields because in both cases, you need to do property Gets on the object (partial date, date-like property bag, or PlainDate). Without custom calendars, you don't need to do the property Gets anymore on PlainDate. But the handling of partial date vs date-like property bag is still very similar. I'm not yet sure of what is going to be the best way to simplify it.

nekevss:
Yeah, the issue I'm running into is that it has to be representable via the type system, if possible, in order to implement in `temporal_rs` over an implementation specific method. So where the specification may have two `JsObject`s that can be handled differently in an abstract op because they're functionally different, we would have to treat as a separate struct in `temporal_rs`. Again, at least if it were to be in `temporal_rs` vs. a specific engine/interpreter implementation. 

Honestly, assuming I'm readying the functionality correctly. This specific split might be easier to reason about than the previous PrepareTemporalFields. Or it could be off base of the intent. I should probably add some unit tests 😅

src/components/mod.rs
```diff
@@ -44,3 +46,90 @@ pub use time::Time;
...
+    Eight,
+    Nine,
+    Ten,
+    Eleven,
+    Twelve,
+    Thirteen,
+}
+
+impl MonthCode {
+    pub fn as_str(&self) -> &str {
```
jasonwilliams, line 71: Will these not change in different calendars? Aren't there more month codes in the chinese calendar? such as https://tc39.es/proposal-intl-era-monthcode/

src/components/calendar.rs
```diff
@@ -622,8 +630,23 @@ impl Calendar {
     /// Provides field keys to be ignored depending on the calendar.
     pub fn field_keys_to_ignore(
         &self,
-        _keys: &[TemporalFieldKey],
+        keys: &[TemporalFieldKey],
     ) -> TemporalResult<Vec<TemporalFieldKey>> {
+        let mut ignored_keys = Vec::default();
```
jasonwilliams, line 635: Should this vec have a minimal capacity? I think right now it will hold 3 keys before reallocating

jasonwilliams (approved):

src/components/mod.rs
nekevss: Yeah, I realized that earlier. I was sort of thinking we could just expand the enum, but that's probably not sufficient either.

src/components/calendar.rs
nekevss: Oooh, potentially. I'll check that!

src/components/mod.rs
jasonwilliams: Maybe this structure is fine for now and we can refactor it to support more in future? What do you think?

src/components/date.rs
```diff
@@ -21,9 +21,40 @@ use std::str::FromStr;
...
 
+// TODO: PrepareTemporalFields expects a type error to be thrown when all partial fields are None/undefined.
+/// A partial Date that may or may not be complete.
+#[derive(Debug, Clone, Copy)]
+pub struct PartialDate {
+    pub(crate) year: Option<i32>,
+    pub(crate) month: Option<i32>,
+    pub(crate) month_code: Option<MonthCode>,
+    pub(crate) day: Option<i32>,
+}
```
ptomato, line 35: What I said in the Temporal meeting about the requirement being "day AND (month OR monthCode) AND (year OR (era AND eraYear))" reminded me that this should probably have optional fields for `era` and `era_year`.

src/components/mod.rs
ptomato: I think if you don't have a "leap" flag in the month code then it probably doesn't make sense to have M13... AFAIK no calendars have an M13, if there is a thirteenth month it's always one of M00L through M12L (or maybe M01L through M11L; it's unclear whether M00L and M12L actually exist)

src/components/date.rs
nekevss: Oh, good point. I had meant to add those in one the update and forgot too.

PR #90, opened 2024-07-31
Add is empty for partialDuration

jasonwilliams:
Changes here are needed for https://github.com/boa-dev/boa/pull/3856

- We need a way of checking if a partialDuration is empty, I don't know if the changes here would work, open to ideas.
- We  need to be able to convert from Finite64 back to f64 so I've added a From impl (for more info see the errors [here](https://github.com/boa-dev/boa/actions/runs/10188786383/job/28185603728?pr=3856):

### DIFF
--- a/src/components/duration.rs
+++ b/src/components/duration.rs
@@ -50,6 +50,15 @@ pub struct PartialDuration {
     pub nanoseconds: Option<FiniteF64>,
 }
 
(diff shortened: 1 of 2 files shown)


nekevss:
I think this is implemented as is_zero and as_inner.

nekevss (approved):

PR #92, opened 2024-08-11
Add `PartialTime` and `PartialDateTime` with corresponding `with` methods.

nekevss:
This continues the work on implementing `with` methods with the approach from #89 by adding the same type of records to `DateTime` and `Time`.

### DIFF
--- a/src/components/datetime.rs
+++ b/src/components/datetime.rs
@@ -279,6 +279,16 @@ impl DateTime {
         partial_datetime: PartialDateTime,
         overflow: Option<ArithmeticOverflow>,
     ) -> TemporalResult<Self> {
+        // Validate that the `PartialDate` is a valid, actionable `PartialDate`
+        if !(partial_datetime.day.is_some()
+            && (partial_datetime.month.is_some() || partial_datetime.month_code.is_some())
+            && (partial_datetime.year.is_some() || (partial_datetime.era.is_some() && partial_datetime.era_year.is_some())))
+        {
+            return Err(TemporalError::r#type()
+                .with_message("A partial date must have at least one defined field."));
+        }
+
+        // Determine the Date from the provided fields.
         let fields = TemporalFields::from(self);
         let partial_fields = TemporalFields::from(PartialDate::from(partial_datetime));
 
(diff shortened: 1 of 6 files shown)


src/components/datetime.rs
```diff
@@ -18,9 +18,68 @@ use tinystr::TinyAsciiStr;
...
     duration::normalized::{NormalizedTimeDuration, RelativeRoundResult},
-    Date, Duration, Time,
+    Date, Duration, MonthCode, PartialDate, PartialTime, Time,
 };
 
+/// A partial DateTime record
+pub struct PartialDateTime {
+    date: PartialDate,
+    time: PartialTime,
+}
```
jedel1043, line 28: Maybe we should just expose the fields as public? having a big method initializer with a lot of parameters seems overkill for essentially an "options" struct that is only passed to an initializer method. It would also make it really easy to set only a single field, since we could call `default` then only initialize the fields we want.

jedel1043 (changes requested):
Overall I agree with this, but I think the API needs to be discussed.

src/components/datetime.rs
nekevss: Yeah, I think that makes sense. I initially went with the above to avoid the multiple parameters of `from_parts`, but then ended up adding it anyways.

nekevss: I went ahead and applied that to all the partials currently completed. starting out with a lot of simplicity is probably the best way to handle it.

src/components/datetime.rs
```diff
@@ -18,9 +18,38 @@ use tinystr::TinyAsciiStr;
...
+    // A potentially set `minute` field.
+    pub minute: Option<i32>,
+    // A potentially set `second` field.
+    pub second: Option<i32>,
+    // A potentially set `millisecond` field.
+    pub millisecond: Option<i32>,
+    // A potentially set `microsecond` field.
+    pub microsecond: Option<i32>,
+    // A potentially set `nanosecond` field.
+    pub nanosecond: Option<i32>,
```
jedel1043, line 50: Thinking that this should reuse `PartialDate` and `PartialTime` instead.

nekevss: Oh, I think I misread the above a bit. I definitely prefer having the internal be `PartialDate` and `PartialTime`, so I'll flip it back

jedel1043 (approved):
Looks great!

PR #94, opened 2024-08-16
Patch for partial records

nekevss:
This PR is a patch for some things I found while doing some implementation of `temporal_rs` in Boa.

The things the PR addresses / patches

- Adds a `DateTime::from_date_and_time`
-  I noticed that the `with` methods weren't checking whether the partials were empty, which would not be valid. This update throws an early error if the partial provided is empty.
- Changed the internal way for constructing a `TemporalFields` from a `PartialDate`. Primarily, after doing the implementation, I don't think having a `From<PartialDate> for TemporalFields` is a good idea. We will still need it internally, but I'm now leaning more towards using the calendar to generate the fields from a partial. Something akin to `calendar::create_fields_from_partial_date(partial_date)`.

### DIFF
--- a/src/fields.rs
+++ b/src/fields.rs
@@ -471,6 +471,40 @@ impl TemporalFields {
 
         Ok(result)
     }
+
+    /// Creates a `TemporalFeild` from a `PartialDate`.
+    pub(crate) fn from_partial_date(partial: &PartialDate) -> Self {
+        let mut bit_map = FieldMap::empty();
+        if partial.year.is_some() {
+            bit_map.set(FieldMap::YEAR, true)
+        };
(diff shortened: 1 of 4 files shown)


src/fields.rs
```diff
@@ -471,6 +471,40 @@ impl TemporalFields {
 
         Ok(result)
     }
+
+    /// Creates a `TemporalFeild` from a `PartialDate`.
```
jedel1043, line 475:
```suggestion
    /// Creates a `TemporalField` from a `PartialDate`.
```

jedel1043 (approved):
Looks good! Only found a typo which doesn't block merging this.

Issue #93, opened 2024-08-11
Should the API for creating the date components be updated?


nekevss:
Currently, to create one of the structs, our general API is `new` with an overflow option. It may be better to have a more "rusty" API; i.e., we lack a `try_new` method.

## General Background:

The Temporal specification provides the option to `Constrain` or `Reject` the creation of an object with `RegulateX` [abstract methods](https://tc39.es/proposal-temporal/#sec-temporal-regulateisodate). To capture this, the current `new` methods all have an `option` parameter that allows choosing the overflow option.

While the current `new` method provides those options, it also means the API doesn't follow typical conventions that may be expected by any Rust users and may also be a little vague.

## Initial Proposal:

Essentially, my thinking is that we should add a `try_new` method and update the API to be a bit more explicit.

- `new(fields..., overflow_option)` => `new_with_overflow_option(fields, overflow_option)`
- Add `try_new(fields ...)` that calls `new_with_overflow_option(fields..., REJECT)`
- Add `new(fields...)` that calls `new_with_overflow_option(fields..., CONSTRAIN)`

That's sort of along the lines of what I was thinking, but I thought I'd bring it up for discussion. I'm not exactly in love with the idea of `new` being constrain by default, but I think it'd be fine as long as it's documented.

Any thoughts?

jasonwilliams:
Are you referring to this? https://github.com/boa-dev/temporal/blob/a7fc946480683074d866490dc75c110688ffdd8e/src/iso.rs#L308-L314

Yeah I guess this could work, I don't have any issue with the idea, I'm not sure it's immediately obvious though so it would need some documentation around it. `new_with_overflow` might be enough (to keep the name shorter). If I just saw `try_new` and `new` on their own I probably wouldn't figure out the `try_new` is "attempt with constrain and fail if it doesn't fit".

>That's sort of along the lines of what I was thinking, but I thought I'd bring it up for discussion. I'm not exactly in love with the idea of new being constrain by default, but I think it'd be fine as long as it's documented.

Yep same, it feels odd, with some documentation we may be OK. I actually don't mind how it is right now, its explicit

nekevss:
> Are you referring to this?

More so around the specific components like `Date`, `Time`, and `DateTime`, so each of these would end up with something akin to the below.

```rust
impl Date {
    // Creates a `Date`, constraining to the nearest valid Date if the values are not in a valid range.
    fn new(
        year: i32,
        month: i32,
        day: i32,
[...]

PR #97, opened 2024-10-19
General API cleanup and adjustments

nekevss:
In general, this completes some various API housekeeping that was needed. Updates the built-ins to align with specification names, changes the exports, implements some of the general API changes to be more Rust-like.

Below are a full list of changes:

  - Updates all component names to align with the Temporal spec (Adds the `Plain` prefix where needed)
  - Initial implementation of the API suggestions from #93
  - Exports each component in the libraries root and adds the partial records to a `partial` mod.

Ideally, now working with the library should look something like the below.

```rust
// previously would have been: 
// use temporal_rs::{components::{calendar::Calendar, Date, PartialDate}, options::ArithmeticOverflow};
use temporal_rs::{Date, Calendar, partial::PartialDate, options::ArithmeticOverflow};

let iso_calendar = Calendar::from_str("iso8601")?;
// Previously would have been Date::new(2020, 10, 19, iso_calendar.clone(), ArithmeticOverflow::Reject)?;
let constructed_date = Date::try_new(2020, 10, 19, iso_calendar.clone())?;
let today = PartialDate {
    year: Some(2020),
    month: Some(10),
    day: Some(19),
    ..Default::default()
};
let date_from_partial = iso_calendar.date_from_partial(&today, ArithmeticOverflow::Reject)?;

assert_eq!(constructed_date, from_partial);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,14 +1,15 @@
-//! The `Temporal` crate is an implementation of ECMAScript's Temporal buitl-in.
+//! The `Temporal` crate is an implementation of ECMAScript's Temporal
+//! buitl-in.
 //!
-//! The crate is being designed with both engine and general use in mind.
+//! The crate is being designed with both engine and general use in
+//! mind.
 //!
-//! IMPORTANT NOTE: Please note that this library is actively being developed and is very
-//! much experimental and NOT STABLE.
+//! [`Temporal`][proposal] is the Stage 3 proposal for ECMAScript that
+//! provides new JS objects and functions for working with dates and
+//! times that fully supports time zones and non-gregorian calendars.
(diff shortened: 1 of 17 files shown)


src/lib.rs
```diff
@@ -1,14 +1,15 @@
-//! The `Temporal` crate is an implementation of ECMAScript's Temporal buitl-in.
+//! The `Temporal` crate is an implementation of ECMAScript's Temporal
+//! buitl-in.
```
jedel1043, line 2:
```suggestion
//! built-in.
```

jedel1043 (approved):
Nice cleanup! I found some typos, but approving preemptively.

PR #98, opened 2024-10-19
Implement `date` macro

nekevss:
This PR implements a `date` macro that would make working with `PartialDate`s ideally a tad more ergonomic. Let me know what you think!


jasonwilliams (approved):
So far so good

nekevss:
Closing as this is fairly stale.

I'm open to the idea of revisiting a macro one day. But I'm not entirely convinced there's an immediate use for it or even general interest.

PR #99, opened 2024-10-25
Stub out tzdb support to unblock `Now` and `ZonedDateTime`

nekevss:
And thus begins the Great Time Zone / `ZonedDateTime` implementation. :sweat_smile: 

This stubs out `tzdb` support by relying solely on the user's zoneinfo files, and should allow us to move forward with implementing and testing `ZonedDateTime` without being blocked by a proper tzdb provider for Boa.

This PR implements:

  - `Now` builtin
  - `FsTzdbProvider` stub
  - Some `ZonedDateTime` APIs along with experimental, non-engine focused APIs.
  
Things of note that are not in this PR:
  
  - Handling POSIX's [proleptic tz format](https://snapshots.sourceware.org/glibc/trunk/2024-06-27_16-55_1719507301/manual/html_node/Proleptic-TZ.html) and non-fat version of the TZDB.
  - Any thing zoneinfo / `zic` related.
  - Tests for more time zone edge cases
  - Many `ZonedDateTime` methods and abstract ops
  - **Windows Support**
  
Let me know what you think :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 authors = ["boa-dev"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/boa-dev/temporal"
-rust-version = "1.79"
+rust-version = "1.80"
 exclude = [
     "docs/*",
     ".github/*",
@@ -26,6 +26,11 @@ num-traits = "0.2.19"
 ixdtf = { version = "0.2.0", features = ["duration"]}
 log = { version = "0.4.0", optional = true }
 
+tzif = { version = "0.2.3", optional = true }
+iana-time-zone = "0.1.61"
 
 [features]
 log = ["dep:log"]
+experimental = ["tzdb"]
+tzdb = ["dep:tzif", "std"]
+std = []
--- a/src/components/tz.rs
+++ b/src/components/tz.rs
@@ -9,7 +9,7 @@ use num_traits::ToPrimitive;
 
 use crate::{components::Instant, iso::IsoDateTime, TemporalError, TemporalResult};
 
-#[cfg(feature = "experimental")]
+#[cfg(all(feature = "tzdb", not(target_os = "windows")))]
 use crate::tzdb::FsTzdbProvider;
 #[cfg(feature = "experimental")]
 use std::sync::{LazyLock, Mutex};
@@ -21,16 +21,16 @@ pub static TZ_PROVIDER: LazyLock<Mutex<FsTzdbProvider>> =
 use super::ZonedDateTime;
 
 pub trait TzProvider {
-    fn check_identifier(&mut self, identifier: &str) -> bool;
+    fn check_identifier(&self, identifier: &str) -> bool;
 
     fn get_named_tz_epoch_nanoseconds(
-        &mut self,
+        &self,
         identifier: &str,
         iso_datetime: IsoDateTime,
     ) -> TemporalResult<Vec<i128>>;
 
     fn get_named_tz_offset_nanoseconds(
-        &mut self,
+        &self,
         identifier: &str,
         epoch_nanoseconds: i128,
     ) -> TemporalResult<i128>;
--- a/src/components/tz.rs
+++ b/src/components/tz.rs
@@ -11,10 +11,10 @@ use crate::{components::Instant, iso::IsoDateTime, TemporalError, TemporalResult
 
 #[cfg(all(feature = "tzdb", not(target_os = "windows")))]
 use crate::tzdb::FsTzdbProvider;
-#[cfg(feature = "experimental")]
+#[cfg(all(feature = "experimental", not(target_os = "windows")))]
 use std::sync::{LazyLock, Mutex};
 
-#[cfg(feature = "experimental")]
+#[cfg(all(feature = "experimental", not(target_os = "windows")))]
 pub static TZ_PROVIDER: LazyLock<Mutex<FsTzdbProvider>> =
     LazyLock::new(|| Mutex::new(FsTzdbProvider::default()));
 
--- a/src/components/tz.rs
+++ b/src/components/tz.rs
@@ -1,46 +1,112 @@
 //! This module implements the Temporal `TimeZone` and components.
 
+use alloc::borrow::ToOwned;
 use alloc::string::String;
 use alloc::vec::Vec;
+use core::{iter::Peekable, str::Chars};
+
 use num_traits::ToPrimitive;
 
-use crate::{
-    components::{calendar::Calendar, Instant, PlainDateTime},
-    TemporalError, TemporalResult,
-};
+use crate::{components::Instant, iso::IsoDateTime, TemporalError, TemporalResult};
+
+#[cfg(feature = "experimental")]
+use crate::tzdb::FsTzdbProvider;
+#[cfg(feature = "experimental")]
+use std::sync::{LazyLock, Mutex};
+
+#[cfg(feature = "experimental")]
+pub static TZ_PROVIDER: LazyLock<Mutex<FsTzdbProvider>> =
+    LazyLock::new(|| Mutex::new(FsTzdbProvider::default()));
+
+use super::ZonedDateTime;
+
+pub trait TzProvider {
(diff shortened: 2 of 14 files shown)


jasonwilliams (commented):
LGTM! 
Good start, nice to have a way forward which will unblock us for now

Cargo.toml
```diff
@@ -25,3 +25,12 @@ bitflags = "2.6.0"
 num-bigint = { version = "0.4.6", features = ["serde"] }
 num-traits = "0.2.19"
 ixdtf = { version = "0.2.0", features = ["duration"]}
+
+
+tzif = "0.2.3"
+iana-time-zone = "0.1.61"
+once_cell = "1.20.2"
```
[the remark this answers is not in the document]
nekevss: Good catch! Forgot to remove it when I moved to `std`.

src/tzdb.rs
```diff
@@ -0,0 +1,364 @@
...
+    self,
+    data::{
+        time::Seconds,
+        tzif::{DataBlock, LocalTimeTypeRecord, TzifData},
+    },
+};
+
+use crate::{components::tz::TzProvider, iso::IsoDateTime, TemporalError, TemporalResult};
+
+const ZONEINFO_DIR: &str = "/usr/share/zoneinfo/";
```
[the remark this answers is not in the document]
nekevss: No, there's not really a windows equivalent, so we will need to package `tzdb` for windows.

That's where we may be able to use `jiff_tzdb` as an intermediate step, but we would need to add steps for resolving from the POSIX tz string as jiff is compiled in the "slim" format of `tzdb` while linux and macos are compiled in "fat"

src/tzdb.rs
```diff
@@ -0,0 +1,457 @@
...
+
+        // The estimated index will be off based on the amount missing
+        // from the lack of offset.
+        //
+        // This means that we may need (idx, idx - 1) or (idx - 1, idx - 2)
+        let record = get_local_record(data_block, estimated_idx);
+        let record_minus_one = get_local_record(data_block, estimated_idx - 1);
+
+        // Potential shift bugs with odd historical transitions?
+        let shift_window = usize::from((record.utoff + record_minus_one.utoff).0.signum() >= 0);
```
jedel1043, line 136: It took me a bit of time to understand that this is just checking if the sum of offsets is positive, so maybe just do that?
```suggestion
let shift_window = usize::from(record.utoff + record_minus_one.utoff >= Seconds(0))
```

src/tzdb.rs
```diff
@@ -0,0 +1,457 @@
...
+        };
+
+        let data_block = self
+            .0
+            .data_block2
+            .as_ref()
+            .expect("binary_search validates that data_block2 exists.");
+
+        let estimated_idx = match b_search_result {
+            Ok(idx) => idx,
```
jedel1043, line 118: Can't we just return the timezone when the binary search finds a corresponding entry (`Ok(idx)`)?

src/tzdb.rs
```diff
@@ -0,0 +1,457 @@
...
+                idx
+            }
+        };
+
+        // The estimated index will be off based on the amount missing
+        // from the lack of offset.
+        //
+        // This means that we may need (idx, idx - 1) or (idx - 1, idx - 2)
+        let record = get_local_record(data_block, estimated_idx);
+        let record_minus_one = get_local_record(data_block, estimated_idx - 1);
```
jedel1043, line 133: I don't think this handles indexing to the transition table correctly. What if the estimated index is one past the end of the table, because the epoch is bigger than any of the transition times?

src/tzdb.rs
```diff
@@ -0,0 +1,457 @@
...
+        self.0
+            .data_block2
+            .as_ref()
+            .map(|b| b.transition_times.binary_search(epoch_seconds))
+    }
+
+    pub fn v2_estimate_tz_pair(
+        &self,
+        seconds: &Seconds,
+    ) -> TemporalResult<Vec<LocalTimeTypeRecord>> {
```
jedel1043, line 103: Can you document what this function does? I _think_ it's just trying to determine the correct timezone data from the current epoch in local time, but it would be good to have it documented.

jedel1043 (commented):
Thank you for all your work! I'm really excited to start playing with the timezone pieces.

Since this is still experimental, we can delay the resolution of my timezone related comments by opening an issue to address that if it would block merging this by a considerable amount.

src/components/tz.rs
```diff
@@ -1,43 +1,109 @@
...
+        identifier: &str,
+        iso_datetime: IsoDateTime,
+    ) -> TemporalResult<Vec<i128>>;
+
+    fn get_named_tz_offset_nanoseconds(
+        &mut self,
+        identifier: &str,
+        epoch_nanoseconds: i128,
+    ) -> TemporalResult<i128>;
+}
```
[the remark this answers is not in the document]
nekevss: I was sort of thinking along the similar lines, but figured I'd go with just for the sake of moving along. I can definitely switch it over.

src/tzdb.rs
```diff
@@ -0,0 +1,457 @@
...
+        }
+    }
+
+    // There are various other ways to search rather than binary_search. See glibc
+    pub fn binary_search(&self, epoch_seconds: &Seconds) -> Option<Result<usize, usize>> {
+        self.0
+            .data_block2
+            .as_ref()
+            .map(|b| b.transition_times.binary_search(epoch_seconds))
+    }
```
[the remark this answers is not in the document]
nekevss: Yeah, I can add that method. I think my primary thoughts were that it would be nice to try and fix this in the `tzif` crate. But in the meantime, adding a method makes sense.

src/tzdb.rs
```diff
@@ -0,0 +1,457 @@
...
+impl FsTzdbProvider {
+    pub fn get(&mut self, identifier: &str) -> TemporalResult<&Tzif> {
+        if !self.cache.contains_key(identifier) {
+            let tzif = Tzif::read_tzif(identifier)?;
+            self.cache.insert(identifier.into(), tzif);
+            self.cache.get(identifier).ok_or(TemporalError::assert())
+        } else {
+            self.cache.get(identifier).ok_or(TemporalError::assert())
+        }
+    }
```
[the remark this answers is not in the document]
nekevss: Huh, I really like this solution, but it looks like it has a lifetime issue around `self.cache.get` having an immutable borrow and `self.cache.entry` borrowing it mutably.

src/tzdb.rs
nekevss: Yeah, if I remember correctly, I think there may have been a bug here, but it should also be solved by #100. If the value is past the transition times, then we need to use the POSIX tz string to resolve.

Ideally, at least the way I am understanding the behavior, we should return the first index if the `seconds` < transition_times[0]. if the value is greater than transition_times[n-1], then the POSIX tz string should be used to resolve the time zone.

src/tzdb.rs
```diff
@@ -90,11 +116,11 @@ impl Tzif {
...
-    pub fn binary_search(&self, epoch_seconds: &Seconds) -> Option<Result<usize, usize>> {
-        self.0
-            .data_block2
-            .as_ref()
-            .map(|b| b.transition_times.binary_search(epoch_seconds))
+    pub fn binary_search(&self, epoch_seconds: &Seconds) -> TemporalResult<Result<usize, usize>> {
+        Ok(self
+            .get_data_block2()?
+            .transition_times
+            .binary_search(epoch_seconds))
```
jedel1043, line 122: My suggestion was more on the lines of removing `binary_search` and just doing this when doing the binary search:
```
let db = self.get_data_block2()?;
let result = db.transition_times.binary_search(epoch_seconds);

nekevss: Ooooooh, we definitely can. [Glibc](https://codebrowser.dev/glibc/glibc/time/tzfile.c.html#__tzfile_compute) does a somewhat gnarly combination of searching that I was intrigued in seeing if it was actually more optimized. Think I was being a bit overly optimistic in separating that out early :sweat_smile:

jedel1043 (approved):
Looks great!

PR #100, opened 2024-10-29
Stub out `tzdb` support for Windows and POSIX tz string

nekevss:
Builds on #99.

This PR adds resolving the POSIX tz string and should allow handling time zones on Windows via `jiff_tzdb` as a source, which should make the stub overall somewhat stable overall until a different solution can be put into place.

### DIFF
--- a/src/tzdb.rs
+++ b/src/tzdb.rs
@@ -34,6 +34,9 @@ use alloc::string::{String, ToString};
 use alloc::{vec, vec::Vec};
 use core::cell::RefCell;
 
+use alloc::string::{String, ToString};
+use alloc::{vec, vec::Vec};
+
 use tzif::{
     self,
     data::{
--- a/src/tzdb.rs
+++ b/src/tzdb.rs
@@ -415,19 +415,21 @@ fn cmp_seconds_to_transitions(
         (TransitionDay::WithLeap(start), TransitionDay::WithLeap(end)) => {
             let day_in_year = utils::epoch_time_to_day_in_year(seconds * 1_000.0) as u16;
             let is_transition = *start == day_in_year || *end == day_in_year;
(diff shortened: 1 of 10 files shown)


jasonwilliams (commented):
I think it might be good to have some description on what the Tzif struct is as there's currently no documentation written above it. I assume it's short for TimeZoneInFo, not sure though. 

I think the same could be said for some of the other new structs in tzdb.rs too like LocalTimeRecord and TransitionTimeSearch etc..

I'm not sure im understanding the slim format difference, the resulting Tzif result looks to be the same size  (552 bytes) whichever format is used. 

Tests look good

nekevss:
Very true. I'll update the docs now.

src/tzdb.rs
```diff
@@ -27,28 +27,114 @@
...
+        }
+    }
+
+    fn from_std(info: &ZoneVariantInfo) -> Self {
+        Self {
+            is_dst: false,
+            offset: -info.offset.0,
+        }
+    }
+}
```
jedel1043, line 78: Maybe `from_standard_time` and `from_daylight_time` instead. The short names are pretty confusing at first glance, especially `from_std` which is commonly used to convert an `std` type into a library type.

nekevss: Yeah, that's probably a fair point. I can switch it over.

nekevss:
> I'm not sure im understanding the slim format difference, the resulting Tzif result looks to be the same size (552 bytes) whichever format is used.

The "slim" vs. "fat" format is laid out by `zic`. Basically, all OSs are compiled using the "fat" format, but if someone does a custom compilation, then a smaller range can be set for which the transition times are computed.

The zic docs can be found [here](https://man7.org/linux/man-pages/man8/zic.8.html). Primarily, `zic`'s bloat and range flags.

jedel1043 (approved):

PR #102, opened 2024-10-30
Implement `ZonedDateTime`'s add and subtract methods

nekevss:
Re: title, depends on #99 and #100

This PR implements `ZonedDateTime`'s add and subtract methods along with related `TimeZone` operations that were needed.

There is also one test to verify the results based off test262.

### DIFF
--- a/src/components/tz.rs
+++ b/src/components/tz.rs
@@ -1,43 +1,109 @@
 //! This module implements the Temporal `TimeZone` and components.
 
-use num_bigint::BigInt;
+use std::{iter::Peekable, str::Chars};
+
 use num_traits::ToPrimitive;
 
-use crate::{
-    components::{calendar::Calendar, Instant, PlainDateTime},
-    TemporalError, TemporalResult,
-};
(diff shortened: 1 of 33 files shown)


src/components/tz.rs
```diff
@@ -109,9 +120,79 @@ impl TimeZone {
         }
     }
 
+    pub fn get_epoch_nanoseconds_for(
+        &self,
+        iso: IsoDateTime,
+        disambiguation: Disambiguation,
+        provider: &mut impl TzProvider,
```
jedel1043, line 127: It doesn't need to be mutable anymore, since we changed the trait methods to be `&self`. Also applies to many other functions.

jedel1043 (changes requested):
Nice work!

src/components/tz.rs
nekevss: Shoot, I meant to change that, but forgot.

jedel1043 (approved):
Looks perfect to me!

PR #103, opened 2024-10-30
Remove num-bigint dependency and rely on primitives

nekevss:
This PR removes the `num_bigint` dependency from the project that was primarily technical debt from the earlier days of `temporal_rs`

### DIFF
--- a/src/components/tz.rs
+++ b/src/components/tz.rs
@@ -2,7 +2,6 @@
 
 use alloc::string::String;
 use alloc::vec::Vec;
-use num_bigint::BigInt;
 use num_traits::ToPrimitive;
 
 use crate::{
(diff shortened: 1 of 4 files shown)


src/components/tz.rs
```diff
@@ -30,13 +29,13 @@ impl TimeZone {
...
     /// Get the offset for this current `TimeZoneSlot`.
-    pub fn get_offset_nanos_for(&self) -> TemporalResult<BigInt> {
+    pub fn get_offset_nanos_for(&self) -> TemporalResult<i128> {
         // 1. Let timeZone be the this value.
         // 2. Perform ? RequireInternalSlot(timeZone, [[InitializedTemporalTimeZone]]).
         // 3. Set instant to ? ToTemporalInstant(instant).
         // 4. If timeZone.[[OffsetMinutes]] is not empty, return 𝔽(timeZone.[[OffsetMinutes]] × (60 × 10^9)).
         if let Some(offset) = &self.offset {
-            return Ok(BigInt::from(i64::from(*offset) * 60_000_000_000i64));
+            return Ok(i128::from(*offset) * 60_000_000_000);
```
jasonwilliams, line 38: We may want to add a comment under the offset property to say its offset minutes from UTC, as it's more descriptively called offsetMinutes in the spec

jasonwilliams (approved):

jedel1043 (approved):
Nice!

PR #110, opened 2024-11-18
Bug fixes and more implementation

jasonwilliams:
- md_record from parse_ixdtf is enough, we shouldn't need to fall back to datetime parsing.
- from_str should lowercase the iso8601 (fixes tests)
- Support parsing the ISO string then getting the calendar from that.
- Support ref year

This is to support changes in https://github.com/boa-dev/boa/pull/4034

### DIFF
--- a/src/components/month_day.rs
+++ b/src/components/month_day.rs
@@ -36,9 +36,11 @@ impl PlainMonthDay {
         day: i32,
         calendar: Calendar,
         overflow: ArithmeticOverflow,
+        ref_year: Option<i32>,
     ) -> TemporalResult<Self> {
+        let ry = ref_year.unwrap_or(1972);
         // 1972 is the first leap year in the Unix epoch (needed to cover all dates)
-        let iso = IsoDate::new_with_overflow(1972, month, day, overflow)?;
+        let iso = IsoDate::new_with_overflow(ry, month, day, overflow)?;
         Ok(Self::new_unchecked(iso, calendar))
     }
 
@@ -54,6 +56,13 @@ impl PlainMonthDay {
     #[must_use]
     pub fn iso_month(&self) -> u8 {
         self.iso.month
+    }
+
+    // returns the iso year value of `MonthDay`.
+    #[inline]
+    #[must_use]
+    pub fn iso_year(&self) -> i32 {
+        self.iso.year
     }
 
     /// Returns the string identifier for the current calendar used.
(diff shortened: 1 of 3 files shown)


src/components/month_day.rs
```diff
@@ -36,9 +36,11 @@ impl PlainMonthDay {
         day: i32,
         calendar: Calendar,
         overflow: ArithmeticOverflow,
+        ref_year: Option<i32>,
```
jedel1043, line 39: I think it makes more sense to put this on the `Calendar` itself, because it makes much more sense to me that the reference year is defined by the calendar.

nekevss: I don't think that the `ref_year` is most likely going to change since the internal fields are ISO fields. The `ref_year` being adjusted via calendar would most likely happen in `month_day_from_partial` if anywhere.

jedel1043: This is not to be able to change the `ref_year`. This change is because it is an implementation-defined thing for each calendar type, and it makes much more sense to me to define that on the calendar and call a method on it than having to carry it over through several calls

nekevss: Hmmmm. Maybe, just to be clear, are you thinking something like a `reference_year` method on `Calendar`? Then the body in this instance would be:

```rust
let ref_year = calendar.reference_year();
let iso = IsoDate::new_with_overflow(ry, month, day, overflow)?;
```

jasonwilliams: I'm now wondering if we need this method at all, it seems in the spec these are mainly created by passing in an isoDate and an calendar, so maybe we need a method that accepts those 2 args instead? And then you use the isoDate to create the overflow?

jedel1043: > I'm now wondering if we need this method at all, it seems in the spec (step 15) these are mainly created by passing in an isoDate and an calendar, so maybe we need a method that accepts those 2 args instead? And then you use the isoDate to create the overflow?

Ohh right, because each calendar would have to convert from their calendar-specific month/day to the equivalent ISO date.

nekevss: I do think that `new`/`new_with_overflow` is definitely needed as it is akin to the [constructor](https://tc39.es/proposal-temporal/#sec-temporal.plainmonthday), but for `ToTemporalMonthDay` that may be able to use `new_unchecked` ... but `month_day_from_partial` returns `PlainMonthDay`, so that should be fine there already.

jedel1043: Got it. We can add the new parameter to `PlainMonthDay`, with the implementation of [`CalendarMonthDayToISOReferenceDate`](https://tc39.es/proposal-temporal/#sec-temporal-calendarmonthdaytoisoreferencedate) still pending.

jedel1043 (approved):

nekevss (approved):

PR #115, opened 2024-11-26
Implement  `PartialZonedDateTime` and `from_partial` and `from_str` for `ZonedDateTime`

nekevss:
Posting this early as a draft as the logic was pretty dense to work through. 

~~This PR still needs some baseline tests added for the methods before it's ready to not be a draft.~~

Update:

Tests have been added! There's also a bug fix in `tzdb.rs`.

Overall, this implements the core logic in temporal_rs for supporting `ToTemporalZonedDateTime` and should unblock implementing `Temporal.ZonedDateTime.from` in Boa.

Overview:

  - Adds `PartialZonedDateTime`
  - Adds `from_partial` and `from_partial_with_provider`
  - Adds `from_str` and `from_str_with_provider`
  - Adds additional tests for the above
  - Fix bug in tzdb.rs `get_local_record`
  - Implements some related abstract ops and moves `IsoDateTime::balance` away from using f64.

### DIFF
--- a/src/components/tz.rs
+++ b/src/components/tz.rs
@@ -17,7 +17,10 @@ use crate::{
 #[cfg(feature = "experimental")]
 use crate::tzdb::FsTzdbProvider;
 #[cfg(feature = "experimental")]
-use std::sync::{LazyLock, Mutex};
+use std::{
+    ops::Deref,
+    sync::{LazyLock, Mutex},
+};
 
 #[cfg(feature = "experimental")]
 pub static TZ_PROVIDER: LazyLock<Mutex<FsTzdbProvider>> =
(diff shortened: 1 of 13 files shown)


src/components/tz.rs
```diff
@@ -7,33 +7,33 @@ use core::{iter::Peekable, str::Chars};
...
-
 pub trait TzProvider {
     fn check_identifier(&self, identifier: &str) -> bool;
 
     fn get_named_tz_epoch_nanoseconds(
         &self,
         identifier: &str,
         iso_datetime: IsoDateTime,
-    ) -> TemporalResult<Vec<i128>>;
+    ) -> TemporalResult<Vec<EpochNanoseconds>>;
```
nekevss, line 36: I noticed when rebasing this PR that `EpochNanoseconds` can (and probably should) be used in more places throughout `temporal_rs`, then I had initially thought. I added some of that work to this PR due to it somewhat being in the scope of work here. For the rest though, I'm thinking it would be best to open up an issue for it and continue looking into using `EpochNanoseconds` in more places for a subsequent PR.

jedel1043: Agreed! Could be a nice first issue for a future contributor.

nekevss: Opened an issue for it 😄

src/components/zoneddatetime.rs
```diff
@@ -1,21 +1,42 @@
...
-use super::{calendar::CalendarDateLike, tz::TzProvider, Duration, PlainDate, PlainDateTime};
-
 #[cfg(feature = "experimental")]
 use crate::components::tz::TZ_PROVIDER;
 #[cfg(feature = "experimental")]
 use std::ops::Deref;
 
+/// A struct representing a partial `ZonedDateTime`.
+pub struct PartialZonedDateTime {
+    /// The `PartialDate` portion of a `PartialDateTime`
```
jasonwilliams, line 30: Is this comment supposed to say "The `PartialDate` portion of a `PartialZonedDateTime`or the `PartialDateTime` is correct?

jasonwilliams (approved):
LGTM!

src/components/zoneddatetime.rs
nekevss: Opps, forgot to type Zoned 😂

PR #116, opened 2024-11-28
Add an `EpochNanosecond` new type

nekevss:
This PR adds an `EpochNanosecond` new type for the internal value of `Instant`. The primary goal of this PR is to clean up the external API around dealing with `Instant` nanoseconds while guaranteeing that the `EpochNanoseconds` value is valid.

### DIFF
--- a/src/components/instant.rs
+++ b/src/components/instant.rs
@@ -15,7 +15,7 @@ use crate::{
     Sign, TemporalError, TemporalResult, TemporalUnwrap,
 };
 
-use num_traits::{Euclid, FromPrimitive, ToPrimitive};
+use num_traits::{Euclid, FromPrimitive};
 
 use super::duration::normalized::NormalizedTimeDuration;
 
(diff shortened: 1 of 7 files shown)


src/components/instant.rs
```diff
@@ -38,17 +78,15 @@ impl Instant {
     ///
     /// Temporal-Proposal equivalent: `AddDurationToOrSubtractDurationFrom`.
     pub(crate) fn add_to_instant(&self, duration: &TimeDuration) -> TemporalResult<Self> {
-        let result = self.epoch_nanoseconds()
+        let current_nanos = self.epoch_nanoseconds() as f64;
```
jasonwilliams, line 81: Why is this cast as an `f64` here?

nekevss: I think this can go either way. It was mostly meant to limit the amount of casts since the addition with the duration fields would require casting at f64 fields to i128.

jasonwilliams (approved):

PR #117, opened 2024-11-28
Add  `ZonedDateTime` calendar accessor methods

nekevss:
This PR is dependent on #115 and will need a rebase once that PR is merged.

This continues to build out methods that `ZonedDateTime` needs for accessor methods.

Overall the features implemented are:

  - All calendar accessor methods
  - An implementation of `HoursInDay` method

### DIFF
--- a/src/components/zoneddatetime.rs
+++ b/src/components/zoneddatetime.rs
@@ -130,7 +130,11 @@ impl ZonedDateTime {
         // 8. Let epochNanoseconds be ? AddZonedDateTime(zonedDateTime.[[EpochNanoseconds]], timeZone, calendar, internalDuration, overflow).
         let epoch_ns = self.add_as_instant(duration, overflow, provider)?;
         // 9. Return ! CreateTemporalZonedDateTime(epochNanoseconds, timeZone, calendar).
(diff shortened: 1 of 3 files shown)


src/components/zoneddatetime.rs
```diff
@@ -358,6 +455,40 @@ impl ZonedDateTime {
...
+impl ZonedDateTime {
+    #[cfg(feature = "experimental")]
+    pub fn hours_in_day(&self) -> TemporalResult<u8> {
+        let provider = TZ_PROVIDER
+            .lock()
+            .map_err(|_| TemporalError::general("Unable to acquire lock"))?;
+        self.hours_in_day_with_provider(provider.deref())
+    }
+
+    pub fn hours_in_day_with_provider(&self, provider: &impl TzProvider) -> TemporalResult<u8> {
```
jedel1043, line 469: Thought: we should probably change the provider argument to `&(impl TzProvider + ?Sized)`, since the current type doesn't allow passing trait objects.

jedel1043 (approved):
Looks good! I just have a small thought that we can address later.

src/components/zoneddatetime.rs
nekevss: Hmmm, true. I'll update in a follow up.

PR #118, opened 2024-11-28
Migrate to `web_time::SystemTime` for `wasm32-unknown-unknown` targets

nekevss:
See title.

This should allow `SystemTime` to work on any wasm targets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,9 @@ tzif = { version = "0.2.3", optional = true }
 jiff-tzdb = { version = "0.1.1", optional = true }
 combine = { version = "4.6.7", optional = true }
 
+# System time feature
+web-time = { version = "1.1.0", optional =  true }
+
 [features]
 log = ["dep:log"]
 experimental = ["tzdb"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -365,6 +365,7 @@ dependencies = [
  "rustc-hash",
  "tinystr",
  "tzif",
+ "web-time",
 ]
 
 [[package]]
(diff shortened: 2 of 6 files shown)


jasonwilliams:
Looking at https://github.com/boa-dev/temporal/pull/118/files#diff-3e4364fa6f0aed7a40e68aa12a41235e57f45378bcd8f0094496a7c1d8be0ed0 It looks like all targets will end up using this crate? Is that a problem or am I missing something?

nekevss:
My understanding from looking at `web_time` is that it's a drop in replacement for `std::time::SystemTime`. If the target isn't wasm, `web_time` just reuses `std::time`

jasonwilliams:
That was my thinking too but wasn’t 100% sure if that’s what’s happening, are we able to confirm it falls back? I guess just by running it right?

Cargo.toml
```diff
@@ -34,6 +34,9 @@ tzif = { version = "0.2.3", optional = true }
 jiff-tzdb = { version = "0.1.1", optional = true }
 combine = { version = "4.6.7", optional = true }
 
+# System time feature
+web-time = { version = "1.1.0", optional =  true }
```
jasonwilliams, line 38: You've marked this as optional but isn't it required? I couldn't test this locally due to the optional being set. If it's designed to fallback to system time I assume the codepath with this will always be used (even on platforms that aren't web)

nekevss: It's optional because `Now` is optional on the `std` feature flag.

jasonwilliams (approved):

PR #121, opened 2024-12-05
Patch the now test per matrix discussion

nekevss:
Previous now test caused sporadic CI failures due to asserting exactly the sleep time. This changes the method of calculation and accounts for sleep being >= to its argument.


jedel1043 (approved):
Nice!

PR #123, opened 2024-12-06
Add now feature flag

nekevss:
This PR adds now feature flag. It fixes a small import bug around the optional import of `web-time` with `sys.rs`.


jedel1043 (approved):
LGTM!

PR #125, opened 2024-12-07
Adjust compilation configuration of tzdb to target_family from target_os

nekevss:
This PR is meant to address issues with boa-dev/boa#4068, which breaks when compiling for WASM targets.

The changes were not made to the tests as I don't believe those will affect compilation.

As a side note, we should probably be testing `temporal_rs` on more targets in CI if possible, especially with adding a FFI on the horizon.


jedel1043 (approved):
Looks good!

PR #126, opened 2024-12-07
Migrate repo to workspace

jedel1043:
This migration enables adding more crates to the repo, like the soon-to-be test262-tester or the temporal_capi FFI.


temporal_rs/README.md
```diff
@@ -0,0 +1,38 @@
+# Temporal in Rust
```
nekevss, line 1: nit: the README.md should probably be in the top level of the crate or a different README.md should probably be made.

nekevss (approved):
I have a couple nits for the workspace, but its not necessarily blocking (1 in a review, the other below here).

I think it would make more sense to label the temporal directory as `temporal` vs. `temporal_rs`.

jedel1043:
@nekevss Almost completely forgot that you can have a root package on a workspace 😅
Changed the structure so that we don't need to have a `temporal` directory, but with a workspace to add crates to.

nekevss (approved):

PR #127, opened 2024-12-07
Remove epochSeconds and epochMicroseconds + adjust epochMillseconds

### LINKED ISSUE #122: Update return type of `Instant::epoch_second` and `Instant::epoch_millisecond`
Currently, these are using `i128`. The return type should be valid for an `i64` value

Seconds: (±8.64 * 10^21) /  10^9
Milliseconds:  (±8.64 * 10^21) /  10^6

nekevss:
This PR updates the epoch* methods on `Instant` and `ZonedDateTime` and closes #122

### DIFF
--- a/src/components/instant.rs
+++ b/src/components/instant.rs
@@ -253,22 +253,10 @@ impl Instant {
         Self::try_new(round_result)
     }
 
-    /// Returns the `epochSeconds` value for this `Instant`.
-    #[must_use]
-    pub fn epoch_seconds(&self) -> i128 {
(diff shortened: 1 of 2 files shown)


jedel1043:
Small question. Why are we removing seconds and microseconds? Is it because it's just a simple multiplication/division by 1000?

nekevss:
Primarily, yes. Secondarily, they were removed from the proposal, and I believe it was precisely because it was a simple multiple/division of 1000 lol

jedel1043 (approved):
LGTM

PR #129, opened 2024-12-09
Add to-x methods and with-x methods to ZonedDateTime

nekevss:
This PR adds support to `ZonedDateTime` for the following methods.

  - ZonedDateTime.prototype.toInstant
  - ZonedDateTime.prototype.toPlainDate
  - ZonedDateTime.prototype.toPlainTime
  - ZonedDateTime.prototype.toPlainDateTime
  - ZonedDateTime.prototype.withTimezone
  - ZonedDateTime.prototype.withCalendar
  - ZonedDateTime.prototype.withTimezone
  - ZonedDateTime.prototype.StartOfDay

### DIFF
--- a/src/components/zoneddatetime.rs
+++ b/src/components/zoneddatetime.rs
@@ -423,7 +423,7 @@ impl ZonedDateTime {
 impl ZonedDateTime {
     /// Creates a new `ZonedDateTime` from the current `ZonedDateTime`
     /// combined with the provided `TimeZone`.
-    pub fn with_plain_time(&self, time: Option<PlainTime>) -> TemporalResult<Self> {
+    pub fn with_plain_time(&self, time: PlainTime) -> TemporalResult<Self> {
         let provider = TZ_PROVIDER
             .lock()
             .map_err(|_| TemporalError::general("Unable to acquire lock"))?;
@@ -710,17 +710,14 @@ impl ZonedDateTime {
     /// combined with the provided `TimeZone`.
     pub fn with_plain_time_and_provider(
         &self,
-        time: Option<PlainTime>,
+        time: PlainTime,
         provider: &impl TzProvider,
     ) -> TemporalResult<Self> {
         let iso = self.tz.get_iso_datetime_for(&self.instant, provider)?;
-        let epoch_ns = if let Some(pt) = time {
-            let result_iso = IsoDateTime::new_unchecked(iso.date, pt.iso);
+        let result_iso = IsoDateTime::new_unchecked(iso.date, time.iso);
+        let epoch_ns =
             self.tz
-                .get_epoch_nanoseconds_for(result_iso, Disambiguation::Compatible, provider)?
-        } else {
-            self.tz.get_start_of_day(&iso.date, provider)?
-        };
+                .get_epoch_nanoseconds_for(result_iso, Disambiguation::Compatible, provider)?;
         Self::try_new(epoch_ns.0, self.calendar.clone(), self.tz.clone())
     }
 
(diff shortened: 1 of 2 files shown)


src/components/zoneddatetime.rs
```diff
@@ -400,6 +420,15 @@ impl ZonedDateTime {
 
 #[cfg(feature = "experimental")]
 impl ZonedDateTime {
+    /// Creates a new `ZonedDateTime` from the current `ZonedDateTime`
+    /// combined with the provided `TimeZone`.
+    pub fn with_plain_time(&self, time: Option<PlainTime>) -> TemporalResult<Self> {
```
jedel1043, line 425: Hmm, does it make sense to take `Option` here? I would assume if someone's calling `with_plain_time`, they would have a specific time to set the datetime to

src/components/zoneddatetime.rs
```diff
@@ -648,6 +705,25 @@ impl ZonedDateTime {
 // ==== Core method implementations ====
 
 impl ZonedDateTime {
+    /// Creates a new `ZonedDateTime` from the current `ZonedDateTime`
+    /// combined with the provided `TimeZone`.
+    pub fn with_plain_time_and_provider(
+        &self,
+        time: Option<PlainTime>,
```
jedel1043, line 712: Same suggestion as my previous comment

src/components/zoneddatetime.rs
nekevss: Yeah, it's a good note.

The [proposal](https://tc39.es/proposal-temporal/#sec-temporal.zoneddatetime.prototype.withplaintime) is fairly clear that the argument for `with_plain_time` is optional and can be undefined. So from a proposal perspective, this is technically correct. But to what degree does it makes sense as an API ...

Hmmmm, sort of thinking out loud. `with_plain_time` is on some level `with_epoch_nanoseconds` where the `IsoDate` of the `ZonedDateTime` is a base.

We could have a `with_epoch_nanosecond` and then reconstruct the functionality in Boa using that method.

From Boa's side, it would probably end up looking like:

```
let zdt = this.downcast_ref ...
let time = args.first().map(|v| to_temporal_time(v)).transpose()?;

let epoch_nanos = if let Some(pt) = time {
    // `as_epoch_nanoseconds` would need to be added to `PlainDateTime`.
    zdt.to_plain_datetime().with_time(pt)?.as_epoch_nanoseconds()
} else {
   zdt.start_of_day(context.tz_provider())?
};

ZonedDateTimeInner::with_epoch_nanoseconds(epoch_nanos)
```

I don't necessarily dislike it ... but it also brings up the question to a point of if it's worthwhile having a `with_epoch_nanoseconds` and `with_instant`.

Thoughts?

nekevss: Or alternatively, we could just support the path where `PlainTime` is available.

```rust
fn with_plain_time(this: &JsValue, args: [JsValue], context: &mut Context) -> JsResult<JsValue> {
    let zdt = this.
        .as_object()
        .and_then(JsObject::downcast_ref::<Self>)
        .ok_or_else( // -- throw error -- //)?;

    let time = args.first().map(|v| to_temporal_time(v)).transpose()?;
    
    if let Some(pt) = time {
        zdt.inner.with_time(pt, context.tz_provider())
    } else {
       zdt.inner.with_instant(zdt.start_of_day(context.tz_provider())?.into())
    }
}
```

jedel1043: Yeah, the second one looks a bit cleaner

jedel1043 (approved):
Nice work!

PR #132, opened 2024-12-12
Fix `epoch_time_to_epoch_year` date equation bug related to `BalanceISODate`

nekevss:
This PR fixes a bug in the current implementation where at the date limits, we were returning incorrect years, and thus balancing wrong as a result. Furthermore, incorrect output was not being verified via `PlainDate::try_new`, so that has been added to `PlainDate::add_date`

NOTE: I am planning on coming back to `epoch_time_to_epoch_year` in the near future after some reading / research. But this does fix the bug.


jedel1043 (approved):
Good fix!

PR #133, opened 2024-12-13
Bump dependencies for `ixdtf`, `tzif`, `icu_calendar`, and `tinystr`.

nekevss:
This bumps the version on the listed dependencies that were closed by dependabot the other week.


Cargo.toml
```diff
@@ -19,15 +19,15 @@ exclude = [
 ]
 
 [workspace.dependencies]
-tinystr = "0.7.6"
-icu_calendar = { version = "1.5.2", default-features = false}
+tinystr = "0.8.0"
+icu_calendar = { version = "2.0.0-beta1", default-features = false}
```
jedel1043, line 23: Do we need a feature from the beta?

nekevss: It's more just an early bump due to what appeared to be incompatibility issues with tinystr 0.8.0

jedel1043 (approved):
LGTM!

PR #137, opened 2024-12-17
Update try_new, new, and new_with_overflow integer type

nekevss:
With the change from #131, there is no longer a need for any of the new methods to use all `i32`s for their arguments. This PR updates `temporal_rs` accordingly.


raskad (approved):

jedel1043 (approved):
Yeah, this is a much better API. Nice work!

PR #140, opened 2024-12-25
Adjust `RelativeTo` according to specification and implementation

nekevss:
These are adjustments to the `RelativeTo` option based off integrating the most recent commits into Boa.

The primary issue is that `RelativeTo` is a specific `to_relative_to` abstract operation (specifically, [GetTemporalRelativeToOption](https://tc39.es/proposal-temporal/#sec-temporal-gettemporalrelativetooption)) that parses a string or property bag obj as a `ZonedDateTime` but has special cases to fall back to a `PlainDate` where that is possible.

This also removes keeping `RelativeTo` as a reference, because it's not really possible to have a reference where the `PlainDateTime`.

### DIFF
--- a/src/components/duration.rs
+++ b/src/components/duration.rs
@@ -628,7 +628,7 @@ impl Duration {
     pub fn round(
         &self,
         options: RoundingOptions,
-        relative_to: Option<RelativeTo<'_>>,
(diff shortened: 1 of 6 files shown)


jedel1043 (approved):
LGTM!

PR #145, opened 2024-12-25
Build out the rest of the Now methods

nekevss:
This PR implements the remaining methods for `Now`'s API.

### DIFF
--- a/src/components/now.rs
+++ b/src/components/now.rs
@@ -10,18 +10,16 @@ use crate::{iso::IsoDateTime, TemporalUnwrap};
 use super::{
     calendar::Calendar,
     timezone::{TimeZone, TzProvider},
-    EpochNanoseconds, Instant, PlainDateTime,
+    EpochNanoseconds, Instant, PlainDate, PlainDateTime, PlainTime, ZonedDateTime,
 };
+
+#[cfg(feature = "experimental")]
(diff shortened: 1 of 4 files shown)


src/components/now.rs
```diff
@@ -10,35 +10,110 @@ use crate::{iso::IsoDateTime, TemporalUnwrap};
...
     calendar::Calendar,
     timezone::{TimeZone, TzProvider},
-    EpochNanoseconds, Instant, PlainDateTime,
+    EpochNanoseconds, Instant, PlainDate, PlainDateTime, PlainTime, ZonedDateTime,
 };
 
+#[cfg(feature = "experimental")]
+use crate::{components::timezone::TZ_PROVIDER, TemporalError};
+#[cfg(feature = "experimental")]
+use std::ops::Deref;
```
jedel1043, line 19: Hmm, I think if using `deref` needs to be gated, it would be better to use the `*` syntax.

jedel1043 (commented):
Looks overall great!! I just have a small suggestion.

src/components/now.rs
nekevss: Yeah, I wasn't the biggest fan of using it at the time. I'm fine removing it.

jedel1043 (approved):
Nice!

PR #146, opened 2024-12-27
Fix RoundingMode::truncation to UnsignedRoundingMode mapping

nekevss:
Currently, the mapping of Rounding Mode to Unsigned Rounding Mode is off for truncation. This fixes the issue and adds a test for it based off the failing test262 test.


jedel1043 (approved):
Nice catch!

PR #147, opened 2025-01-06
Implement Neri-Schneider calculations

nekevss:
This PR changes the current date calculations to one's layed out by [Neri-Schneider's Euclidean affine functions paper](https://onlinelibrary.wiley.com/doi/full/10.1002/spe.3172).

There is still some documentation that needs to be completed before this is ready to merge. I was also hoping to double check the assembly but I'm sort of viewing that as part of the documentation. In the meantime, I'm posting this as a draft for feedback.


cassioneri:
@nekevss Thanks for your interest in our work and for this implementation. Lorenz and I are always glad to see our algorithms to be implemented by others in different computer systems.

I took the liberty to inspect the assembly generated by your implementation through [compiler explorer](https://godbolt.org). Here is the [link](https://godbolt.org/z/PaPK79rTd). It looks quite nice already but it can be improved a bit further.

In the paper we define a constant `s` which shifts the interval of supported dates. More specifically, we set `s=82` to comply with the C++ Standard that requires support for years in `[-32767, 32767]`. This was just an example and, as you rightly figure out, can be changed to support different ranges. However, in your code `s` is a variable to allow users to set their own preference and for Temporal you set `s=680`. Using a variable might have performance costs as indicated by comparing the assembly in the link above to the one obtained when `s` is a constant as shown [here](https://godbolt.org/z/3frT7dfEq). (Notice the definition of `const CYCLE_SHIFTS` at line 34.)

IMHO, it's better to define `s` as a constant and the best value for it is `3670`. This choice places `1970/Jan/01` roughly in  the middle of the interval of supported dates, namely, `[-1468000/Mar/01, 1471805/Jun/05]`. Unfortunately, this idea came to my mind only after the paper was published. I used this value in my implementation for SpiderMonkey. You can see it [here](https://hg.mozilla.org/integration/autoland/rev/54ebf8bd2e11#l3.70) with a lengthy comment explaining this choice.

src/neri_schneider.rs
```diff
@@ -0,0 +1,256 @@
...
+pub const fn days_in_century(rata_die: u32) -> u32 {
+    century_rem(rata_die).div_euclid(4)
+}
+
+/// returns Y, N_y AKA, year and day_of_year
+const fn second_equations(rata_die: u32) -> (u32, u32) {
+    let (century, rem) = first_equations(rata_die);
+    let n_two = rem | 3;
+    let year_of_century = (376_287_347 * n_two as u64).div_euclid(TWO_POWER_THIRTY_NINE) as u32;
+    let day_of_year = (n_two - 1461 * year_of_century).div_euclid(4);
```
cassioneri, line 110: As explained in Section 8.2 of our paper, although the second column of algorithm 3 is written as
```
Z = N_2 / 1461
N_Y = N_2 % 1461 / 4
```
The code usually generated by optimiser compilers actually calculates
```
Z =  376_287_347 * N_2 / 2^39
N_Y = (N_2 - 1461 * Z) / 4
```
Hence, the code as written here, just replicates optimisations that compilers usually do for us. This can be seen [here](https://godbolt.org/z/zqq8Pr6xG). (Fair enough, the assembly for the code in this PR saves one `mov` instruction but this is not a big deal.)

The section goes on and explains that the second column of algorithm 5 is a better alternative:
```
P_2 = 2_939_745 * N_2
Z = P2 / 2^32
N_Y = P_2 % 2^32 / 2_939_745 / 4
```
Although the assembly for the above and the PR's code use a quite similar list of instructions ([here](https://godbolt.org/z/b34Tjddhv)), the main advantage of the above is that it breaks the dependency of `N_Y` on `Z` which allows modern CPUs to start calculating `N_Y` in parallel, before `Z` is obtained.

Hence, I suggest:
```suggestion
    let p2 = 2_939_745 * n_two as u64;
    let year_of_century = p2.div_euclid(TWO_POWER_THIRTY_TWO) as u32;
    let day_of_year = p2.rem_euclid(TWO_POWER_THIRTY_TWO).div_euclid(2_939_745).div_euclid(4) as u32;
```

Also note that `p2.rem_euclid(TWO_POWER_THIRTY_TWO)` is equivalent as `p2 as u32` which is shorter but has no impact whatsoever in the generated assembly. The choice between the two is a matter of taste.

src/neri_schneider.rs
jedel1043, line 1: I'm a bit skeptical about exposing this as a public module. I would expect this to be just internal details of our `utils` module.

jedel1043 (changes requested):
Really impressive work! I'm excited to see that we're finally getting rid of those pesky `f64` in our calculations 😅

src/neri_schneider.rs
```diff
@@ -0,0 +1,275 @@
...
+pub const DAYS_IN_A_400Y_CYCLE: u32 = 146_097;
+
+const TWO_POWER_THIRTY_NINE: u64 = 549_755_813_888; // 2^39 constant
+const TWO_POWER_THIRTY_TWO: u64 = 4_294_967_296; // 2^32 constant
+const TWO_POWER_SIXTEEN: u32 = 65_536; // 2^16 constant
+const DAYS_IN_GREGORIAN_CYCLE: i32 = DAYS_IN_A_400Y_CYCLE as i32;
+const SHIFT_CONSTANT: i32 = 3670;
+
+/// Calculate Rata Die value from gregorian
+pub fn epoch_days_from_gregorian_date(year: i32, month: i32, day: i32) -> i32 {
```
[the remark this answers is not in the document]
nekevss: Ooooh, good point. Actually, I think this might have been a bug on the implementation side connected to IsoDate::balance

src/neri_schneider.rs
jedel1043: This is still exposed as a public (crate) module. I think the best option here is to have `neri_schneider` be a submodule of `utils`, then just avoid exposing implementation details to the rest of the library.

nekevss: Ah, I misread that for some reason. Making it a submodule of `utils` sounds fine to me.

jedel1043 (approved):
Really nice implementation! There are still some details that we could definitely improve — explain what "rata die",  "computational calendar", "first equation", etc. mean, maybe change some names to be more descriptive — but I think this is good enough that we can merge it and then iteratively improve on it.

PR #153, opened 2025-01-07
Implement support for `to_string` and implement `PlainDate::to_string`

nekevss:
This PR implements the underlying code to implement `Display`, `to_string`, and `to_ixdtf_string` methods for nearly all the main built-ins aside from `Duration`, `PlainYearMonth`, and `PlainMonthDay`.

The majority of the backing implementation will most likely need to eventually be moved into the `ixdtf` crate, but until that is complete and worked, this should allow us to move forward with the implementation and testing in Boa against test262.

### DIFF
--- a/src/parsers.rs
+++ b/src/parsers.rs
@@ -118,7 +118,6 @@ fn write_nanosecond<W: core::fmt::Write + ?Sized>(
 pub fn u32_to_digits(mut value: u32) -> ([u8; 9], usize) {
     let mut output = [0; 9];
     let mut precision = 0;
-    // let mut precision_check = 0;
     let mut i = 9;
     while i != 0 {
         let v = (value % 10) as u8;
--- a/src/parsers.rs
+++ b/src/parsers.rs
@@ -9,7 +9,7 @@ use ixdtf::parsers::{
     records::{Annotation, DateRecord, IxdtfParseRecord, TimeRecord, UtcOffsetRecordOrZ},
     IxdtfParser,
 };
-use writeable::{impl_display_with_writeable, Writeable};
+use writeable::{impl_display_with_writeable, LengthHint, Writeable};
 
 // TODO: Move `Writeable` functionality to `ixdtf` crate
 
@@ -47,9 +47,22 @@ impl Writeable for FormattableTime {
             return Ok(());
         }
         sink.write_char('.')?;
-        write_nanosecond(self.nanosecond, self.precision, sink)?;
+        write_nanosecond(self.nanosecond, self.precision, sink)
+    }
 
-        Ok(())
+    fn writeable_length_hint(&self) -> LengthHint {
+        let sep = self.include_sep as usize;
+        if self.precision == Precision::Minute {
+            return LengthHint::exact(4 + sep);
+        }
+        let time_base = 6 + (sep * 2);
+        if self.nanosecond == 0 {
+            return LengthHint::exact(time_base);
+        }
+        if let Precision::Digit(d) = self.precision {
+            return LengthHint::exact(time_base + 1 + d as usize);
+        }
+        LengthHint::between(time_base + 2, time_base + 10)
     }
 }
 
@@ -71,6 +84,13 @@ impl Writeable for FormattableUtcOffset {
         match &self.offset {
             UtcOffset::Z => sink.write_char('Z'),
             UtcOffset::Offset(offset) => offset.write_to(sink),
+        }
+    }
+
+    fn writeable_length_hint(&self) -> LengthHint {
+        match &self.offset {
+            UtcOffset::Z => LengthHint::exact(1),
+            UtcOffset::Offset(o) => o.writeable_length_hint(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/parsers.rs
```diff
@@ -1,13 +1,276 @@
...
+        }
+        write_padded_u8(self.second, sink)?;
+        if self.nanosecond == 0 {
+            return Ok(());
+        }
+        sink.write_char('.')?;
+        write_nanosecond(self.nanosecond, self.precision, sink)?;
+
+        Ok(())
+    }
```
sffc, line 53: Suggestion: your `impl Writeable` may benefit from a `write_len` function since it looks fairly cheap to pre-compute. You could add one later.

src/parsers.rs
```diff
@@ -1,13 +1,276 @@
...
+}
+
+fn write_nanosecond<W: core::fmt::Write + ?Sized>(
+    nanoseconds: u32,
+    precision: Precision,
+    sink: &mut W,
+) -> core::fmt::Result {
+    if nanoseconds > 1_000_000_000 {
+        return Err(core::fmt::Error);
+    }
```
sffc, line 93: Nit: You should avoid returning `core::fmt::Error` because it gets turned into an unhandled `panic!` in most call sites. You should try and ensure that your nanoseconds are in range by adding invariants to the type being formatted, instead of returning an error in the Writeable impl.

If you need to return errors from the Writeable impl, consider using TryWriteable instead.

src/parsers.rs
```diff
@@ -1,13 +1,276 @@
...
+    sink: &mut W,
+) -> core::fmt::Result {
+    let mut divisor = 100_000_000;
+    while divisor >= 1 && nanoseconds != 0 {
+        (nanoseconds / divisor).write_to(sink)?;
+        nanoseconds %= divisor;
+        divisor /= 10;
+    }
+
+    Ok(())
```
sffc, line 111: Observation: The compiler can often optimize out a small fixed operation; for example, I think it changes `/10` from an expensive `div` instruction to some clever bit shifting stuff. However, since the `divisor` is changing on each iteration of the loop, the compiler is less likely to do this optimization and may leave around an actual `div` instruction in the assembly.

You should consider making a little helper that returns `[u8; 9]` with the ASCII bytes of a padded 9-digit number, which you can fill in backwards to get the `/10` optimization, and then write the subslice that you need from it into the sink. You can use it in all of these places where you need to write a padded number to a string. Unfortunately this may involve a small amount of unsafe code since you need to assert that the temporary `[u8; 9]` buffer is UTF-8. However, hopefully once [`core::ascii`](https://doc.rust-lang.org/beta/core/ascii/index.html) gets stabilized, you will make such code safe.

I may even support adding the helper function to `writeable` itself, but you should first develop it here.

sffc (commented):
Praise: Thanks for using `writeable` :)

src/parsers.rs
```diff
@@ -1,13 +1,335 @@
...
+    };
+    // SAFETY: Index must be within a valid range of 0..=9 and is
+    // valid aschii digit chars.
+    sink.write_str(unsafe { core::str::from_utf8_unchecked(slice) })
+}
+
+pub fn write_u32_to_ascii_digits(mut value: u32) -> ([u8; 9], usize) {
+    let mut output = [0; 9];
+    let mut precision = 0;
+    // let mut precision_check = 0;
```
raskad, line 123:
```suggestion
```

raskad (approved):
Looks good

src/parsers.rs
```diff
@@ -150,3 +472,116 @@ pub(crate) fn parse_time(source: &str) -> TemporalResult<TimeRecord> {
...
+            time: FormattableTime {
+                hour: 4,
+                minute: 0,
+                second: 0,
+                nanosecond: 0,
+                precision: Precision::Minute,
+                include_sep: true,
+            },
+        };
+        assert_eq!(offset.to_string(), "+04:00");
```
sffc, line 495: Nit: Use `assert_writeable_eq` instead of `assert_eq` in order to test the Writeable impl more thoroughly (such as the correctness of `writeable_length_hint`)

src/parsers.rs
```diff
@@ -1,13 +1,335 @@
...
+        }
+    }
+}
+
+fn write_padded_u8<W: core::fmt::Write + ?Sized>(num: u8, sink: &mut W) -> core::fmt::Result {
+    if num < 10 {
+        sink.write_char('0')?;
+    }
+    num.write_to(sink)
+}
```
sffc, line 103: Suggestion: I like how you use `impl Writeable for u8` here. I think you can do that with the u32 as well.

Something like:

```rust
let mut max = 1e9;
while max > nanoseconds {
    sink.write_char('0')?;
    max /= 10;
}
nanoseconds.write_to(sink)?;
```

And then you don't need any unsafe code.

nekevss: Does that fix output the nanoseconds according to what would be expected? If I'm reading that write, it would left pad nanoseconds to always be 9 digits, but the concern with this is to trim trailing zeros.

nanoseconds = 123_000_000, should only print .123 or the desired precision, i.e. precision = 5 would be .12300, but this would print 123000000

The idea for the use of unsafe in the function was to try and optimize out the `div` instruction from the below, which I think would be the equivalent to the above.

```
   let mut divisor = 1_000_000_000;
    while divisor >= 1 && nanoseconds != 0 {
        (nanoseconds / divisor).write_to(sink)?;
        nanoseconds %= divisor;
        divisor /= 10;
    }
```

nekevss: Actually, nevermind. We can just write the digits manually as `u8`s

sffc (approved):

PR #156, opened 2025-01-12
Fix edge case for disambiguating `ZonedDateTime`s on DSTs skipping midnight

jedel1043:
Fixes part of the test [intl402/Temporal/ZonedDateTime/from/dst-skipped-cross-midnight.js](https://github.com/tc39/test262/blob/d9b10790bc4bb5b3e1aa895f11cbd2d31a5ec743/test/intl402/Temporal/ZonedDateTime/from/dst-skipped-cross-midnight.js). This doesn't fix correctly resolving the start of the day for days that don't start at 00:00, since that'll require more advanced logic.

### DIFF
--- a/src/components/zoneddatetime.rs
+++ b/src/components/zoneddatetime.rs
@@ -29,6 +29,7 @@ use crate::{
 use crate::components::timezone::TZ_PROVIDER;
 
 /// A struct representing a partial `ZonedDateTime`.
+#[derive(Debug, Default, Clone, PartialEq)]
 pub struct PartialZonedDateTime {
     /// The `PartialDate` portion of a `PartialZonedDateTime`
     pub date: PartialDate,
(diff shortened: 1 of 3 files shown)


src/components/zoneddatetime.rs
```diff
@@ -1344,4 +1344,19 @@ mod tests {
         );
         assert!(result.is_ok());
     }
+
+    #[test]
+    // https://github.com/tc39/test262/blob/d9b10790bc4bb5b3e1aa895f11cbd2d31a5ec743/test/intl402/Temporal/ZonedDateTime/from/dst-skipped-cross-midnight.js
+    fn dst_skipped_cross_midnight() {
+        let provider = &FsTzdbProvider::default();
+        let midnight_disambiguated = ZonedDateTime::from_str_with_provider(
+            "1919-03-31T00[America/Toronto]",
```
nekevss, line 1353: Ah, shoot. Toronto 1919 🫠

src/tzdb.rs
```diff
@@ -557,14 +558,14 @@ impl TzProvider for FsTzdbProvider {
             LocalTimeRecordResult::Empty => Vec::default(),
             LocalTimeRecordResult::Single(r) => {
                 let epoch_ns =
-                    EpochNanoseconds::try_from(epoch_nanos.0 + seconds_to_nanoseconds(r.offset))?;
+                    EpochNanoseconds::try_from(epoch_nanos.0 - seconds_to_nanoseconds(r.offset))?;
```
nekevss, line 561: Huh, is this correct? I think the tests below due assert that the time zones are not inverted like in PosixTzString, so I'm pretty sure the sign needs to be minus (I could definitely be off here). Ex: EST is -04:00, to move the local back to UTC you local - -04:00.

jedel1043: Yes. The input is a local time, and the output is an offset from UT. Hence, since `local = UT + offset`, then `local - offset = UT`

nekevss: OH wait, I had put plus oh god :rofl:

nekevss (approved):

PR #157, opened 2025-01-13
Fixing panics in test262 when running in debug mode

nekevss:
This PR fixes nearly all of the panics that occur when running temporal_rs through the test262 suite using Boa as noted by @jedel1043.


jedel1043 (approved):
Nice!

PR #158, opened 2025-01-13
Add an MSRV check to CI

nekevss:
This adds an MSRV check to CI so that changes to `temporal_rs` are hopefully always inline with Boa's MSRV.


jedel1043 (approved):
Good to have a msrv check!

PR #159, opened 2025-01-13
Implement correct resolving of `getStartOfDay`

jedel1043:
Still unimplemented for the specific case of a month-week-day posix string.

### DIFF
--- a/src/tzdb.rs
+++ b/src/tzdb.rs
@@ -208,10 +208,25 @@ impl Tzif {
 
     pub fn get(&self, epoch_seconds: &Seconds) -> TemporalResult<TimeZoneOffset> {
         let db = self.get_data_block2()?;
+
         let result = db.transition_times.binary_search(epoch_seconds);
 
         match result {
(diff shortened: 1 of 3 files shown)


src/tzdb.rs
```diff
@@ -69,14 +70,14 @@ impl LocalTimeRecord {
...
             is_dst: true,
-            offset: -info.offset.0,
+            offset: info.offset.0,
         }
     }
 
     fn from_standard_time(info: &ZoneVariantInfo) -> Self {
         Self {
             is_dst: false,
-            offset: -info.offset.0,
```
nekevss, line 79: The `ZoneVariantInfo` type was used primarily on the POSIX tz string. Inverting here is correct, because the time zones are inverted in the POSIX tz format. Australia is at -11 and EST is +5

jedel1043: Got it.

nekevss (approved):
Looks great!

PR #161, opened 2025-01-13
Add to_string functionality for timezone identifer

nekevss:
Forgot to implement this alongside all the other to string functionality.


raskad (approved):
Looks good

PR #162, opened 2025-01-14
Bug fixes to address test failures + removing unused API

nekevss:
This PR fixes both an assertion bug from caused by `IncrementRounder`'s as_positive path. Since this was the only use case in the specification that seemed to be using the "as_positive" path, I completely removed the methods.

This also fixes some other bugs caused from using non_euclidean operations in `IsoDateTime::from_epoch_nanos`


jedel1043 (approved):
Nice bug fixing! I like the new name for `from_potentially_negative_parts`. `from_signed_num` is short and simple.

PR #168, opened 2025-01-17
API cleanup, visibility updates, and tech debt cleanup

nekevss:
This PR addresses some issues brought up in #166.

The general updates are as follows:

  - `CalendarDateLike` removed in favor of using `IsoDate` directly.
  - `CalendarFieldsType` removed
  - `GetTemporalCalendar` removed
  - `DateDuration` and `TimeDuration` exported from root with the other builtins
  - `TzProvider` renamed to `TimeZoneProvider`


jedel1043 (approved):
API looks much cleaner now!

PR #169, opened 2025-01-18
Restructure project to separate core provider APIs from non-provider APIs

nekevss:
This PR lays out the general concept laid out in #165.

Primarily, this PR splits the core Temporal functionality that can be used by engines into the `builtins::core` module and then builds out wrappers for nearly all of the core builtins (Currently only `PlainYearMonth` and `PlainMonthDay` have not had wrappers implemented) that can be used directly from Rust in the `builtins::native` module.

Ideally, this provides a better experience for using the library natively from Rust (while also giving us a cleaner API to write tests with), but still allows `with_provider` API to be used for engines and/or anyone else who wants to implement their own `TimeZoneProvider`.


jasonwilliams:
I don't think we need to be as detailed but it would be nice to generate an `architecture.md` file similar to https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/architecture.md which can then explain the dir structure.

jedel1043 (approved):
Completely agree with all the changes. I just have a small typo fix and then we can merge this :)

PR #170, opened 2025-01-19
Implement `ZonedDateTime::since` and `ZonedDateTime::until`

nekevss:
This PR implements the `since` and `until` methods for `ZonedDateTime`.

It also has a host of various bug fixes that I came across from implementing the methods in Boa and running the test suite. As a result, each method has a ~80% conformance rating (since 72/90 and until 70/88).

### DIFF
--- a/src/components/zoneddatetime.rs
+++ b/src/components/zoneddatetime.rs
@@ -13,9 +13,7 @@ use crate::{
     },
     iso::{IsoDate, IsoDateTime, IsoTime},
     options::{
-        ArithmeticOverflow, Disambiguation, DisplayCalendar, DisplayOffset, DisplayTimeZone,
-        OffsetDisambiguation, ResolvedRoundingOptions, RoundingIncrement, TemporalRoundingMode,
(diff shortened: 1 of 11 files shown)


src/components/zoneddatetime.rs
```diff
@@ -249,6 +251,74 @@ impl ZonedDateTime {
...
+    /// `temporal_rs` equivalent to `DifferenceTemporalZonedDateTime`.
+    pub(crate) fn diff_internal_with_provider(
+        &self,
+        op: DifferenceOperation,
+        other: &Self,
+        options: DifferenceSettings,
+        provider: &impl TimeZoneProvider,
+    ) -> TemporalResult<Duration> {
+        // NOTE: for order of operations, this should be asserted prior to this point
+        // by implementors any engine implementors, but asserting out of caution.
```
jasonwilliams, line 264:
```suggestion
        // by any engine implementors, but asserting out of caution.
```

jasonwilliams (approved):
LGTM

PR #173, opened 2025-01-21
Fix `Duration` parsing not returning a range error.

nekevss:
Per title, parsing should return a range error and not a general error.


jedel1043 (approved):
Nice catch!

PR #180, opened 2025-01-22
Revert "Restructure project to separate core provider APIs from non-provider APIs"

nekevss:
Reverts boa-dev/temporal#169.

As mentioned in #165, this should be handled differently then this approach.


jedel1043:
I don't think splitting them between modules is bad; it's easier to handle feature-gating and maintainance. IMO we should preserve the current "core" and "native" structure (maybe with some further bikeshed, not that important), but use impl blocks instead of wrappers.

PR #182, opened 2025-01-24
Cleanup `Now` API in favor of system defined implementations

nekevss:
~~This adds a `SystemHooks` trait for providing system getters for system time and system time zone. It also adjusts the "now" feature to a "sys" feature flag for `DefaultSystemHooks`.~~

This PR provides more generalized methods that can be used in combination with a system hooks.

### DIFF
--- a/src/builtins/core/now.rs
+++ b/src/builtins/core/now.rs
@@ -109,7 +109,7 @@ impl Now {
     ///   1. Resolve user input `TimeZone` with the `SystemTimeZone`.
     ///   2. Get the `SystemNanoseconds`
     ///
-    /// For an example implementation see [`Self::plain_date_time_iso`]
+    /// For an example implementation see [`Self::plain_datetime_iso`]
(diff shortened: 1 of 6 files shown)


src/builtins/core/now.rs
```diff
@@ -16,24 +17,61 @@ use super::{
 pub struct Now;
 
 impl Now {
-    /// Returns the current instant
-    pub fn instant() -> TemporalResult<Instant> {
-        system_instant()
+    pub fn instant_with_epoch_nanoseconds(epoch_nanoseconds: EpochNanoseconds) -> Instant {
```
raskad, line 20: Missing doc comment for a pub fn

raskad (approved):
Looks good

src/builtins/core/now.rs
nekevss: Shoot good catch. I'll fix that up

PR #183, opened 2025-01-25
Reimplement Unit Group with different approach

nekevss:
This PR reimplements the UnitGroup enum with a different approach to validation. Previously this was implemented, but it broke quite a few tests in Boa. I think that was partially due to a flawed implementation and a host of bugs that were previously present.

Those appear to have all been fixed with recent bug changes, and this reimplementation is primarily to address the broken tests in boa-dev/boa#4142 that require unit groups to be implemented.

The tests that the unit groups specifically address specifically are tests like: [largestunit-invalid-string.js](https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Instant/prototype/since/largestunit-invalid-string.js).


jedel1043 (approved):
Looks good!

PR #186, opened 2025-01-27
Temporal duration compare

### LINKED ISSUE #154: Implement `compare` method for `Duration`
Related to ongoing work for #11.

The `compare` method for `Duration` needs an implementation according to spec in order to implement `Temporal.Duration.compare`.

sffc:
Fixes #154

Wrote this along with @lockels, @neelzee, @sebastianjacmatt, @magnus-fjeldstad, @HenrikTennebekk.

Please see question regarding the return type of DateDurationDays.


src/builtins/core/duration.rs
```diff
@@ -277,6 +277,65 @@ impl Duration {
     pub fn is_time_within_range(&self) -> bool {
         self.time.is_within_range()
     }
+
+    /// Returns the ordering between the two durations.
+    /// 
+    /// `7.2.3 Temporal.Duration.compare ( one, two [ , options ] )`
+    #[must_use]
+    pub fn compare(
```
nekevss, line 285: question: Should compare take `&self` for the first `Duration` parameter or mimic the specification and have it be a standalone function on `Duration`?

This is more a question of what is the best way to translate `compare` in Rust vs. JavaScript. In the literal sense to the specification, this is implemented exactly as it should be. But in Rust where `PartialOrd` exists as a trait, it's common to have `one.cmp(&two)` conceptually. So is it best for this to be a method similar to `PartialOrd`, e.g. `one.compare(&two, None)?`?

I've tended to prefer this approach, mostly, because I think it's nice to adhere to similarly structured syntax. But I could be convinced to this current implementation if others think it's bad to overload the concept of `one.cmp(&two)`. The most important thing to be consistent across similar methods and functions in `temporal_rs` for the builtins.

src/builtins/core/duration.rs
```diff
@@ -77,7 +77,7 @@ impl PartialDuration {
 /// `Duration` is made up of a `DateDuration` and `TimeDuration` as primarily
 /// defined by Abtract Operation 7.5.1-5.
 #[non_exhaustive]
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Clone, Copy, Default, PartialEq, PartialOrd)]
```
nekevss, line 80: issue: `Duration` should not implement `PartialOrd` and potentially not `PartialEq`.

The `compare` method supersedes the `cmp` impl from `PartialOrd`, implementing them both could lead to confusion. `PartialEq` is not necessarily needed by the specification. It doesn't define an `equals` method on `Duration`, but `PartialEq` is incredibly useful for testing purposes.

src/builtins/core/duration.rs
```diff
@@ -277,6 +277,65 @@ impl Duration {
...
+                // a. Let timeZone be zonedRelativeTo.[[TimeZone]].
+                // b. Let calendar be zonedRelativeTo.[[Calendar]].
+                // c. Let after1 be ? AddZonedDateTime(zonedRelativeTo.[[EpochNanoseconds]], timeZone, calendar, duration1, constrain).
+                // d. Let after2 be ? AddZonedDateTime(zonedRelativeTo.[[EpochNanoseconds]], timeZone, calendar, duration2, constrain).
+                let after1 = zdt.add(one, Some(ArithmeticOverflow::Constrain))?;
+                let after2 = zdt.add(two, Some(ArithmeticOverflow::Constrain))?;
+                // e. If after1 > after2, return 1𝔽.
+                // f. If after1 < after2, return -1𝔽.
+                // g. Return +0𝔽.
+                return Ok(after1.cmp(&after2));
```
nekevss, line 311: note: this will probably need to be updated on a rebase / merge to main due to #175.

src/builtins/core/duration/date.rs
```diff
@@ -105,4 +108,60 @@ impl DateDuration {
...
+        );
+        // 6. Let yearsMonthsWeeksInDays be epochDays2 - epochDays1.
+        let ymd_in_days = epoch_days_2 - epoch_days_1;
+        // 7. Return dateDuration.[[Days]] + yearsMonthsWeeksInDays.
+        Ok(self.days.as_integer_if_integral::<i64>()?
+            + i64::try_from(ymd_in_days).map_err(|_| TemporalError::range())?)
+    }
+
+    /// AdjustDateDurationRecord
+    pub fn adjust(
```
nekevss, line 149: suggestion: `pub(crate)`

Since this is an abstract op, it would be best to keep these `pub(crate)` until otherwise needed.

nekevss (commented):
Thanks for contributing! This is looking really great overall!

src/builtins/core/duration.rs
nekevss: For reference on the above comment:

https://github.com/boa-dev/temporal/blob/main/src/builtins/core/zoneddatetime.rs#L434

src/builtins/core/duration.rs
```diff
@@ -277,6 +277,73 @@ impl Duration {
     pub fn is_time_within_range(&self) -> bool {
         self.time.is_within_range()
     }
+
+    /// Compares self with other on a field by field basis.
+    fn cmp_raw(&self, other: &Self) -> Option<Ordering> {
```
lockels, line 282: Should this method be moved to a different section of the code that's private with pub(crate)? This is not part of the public API, but instead a helper method for the actual implementation of ``compare``

src/builtins/compiled/duration.rs
```diff
@@ -23,4 +25,20 @@ impl Duration {
...
             .map(Into::into)
     }
+
+    /// Returns the ordering between two [`Duration`], takes an optional
+    /// [`RelativeTo`]
+    ///
+    /// Enable with the `compiled_data` feature flag.
+    pub fn compare(
+        &self,
+        two: &Duration,
```
nekevss, line 35: nit: as this is a public API, I think I'd prefer `other` here over `two`.

src/builtins/core/duration.rs
nekevss: Yes! Good question / catch! Preferably it should live in a different block.

src/options.rs
```diff
@@ -424,6 +424,13 @@ impl TemporalUnit {
         matches!(self, Year | Month | Week)
     }
 
+    #[inline]
+    #[must_use]
+    pub fn is_date_unit(&self) -> bool {
```
nekevss, line 429: this is actually a long needed addition lol

src/builtins/core/duration.rs
lockels: Alright:) Which block exactly is the best place to put it? Private creation methods? Or maybe outside the any of the impl blocks? None of them make sense as a natural spot to place this function to me

lockels: oh, that's true! thanks a lot, i'll remove the redundant function and just inline it instead

src/builtins/core/duration/date.rs
```diff
@@ -105,4 +108,59 @@ impl DateDuration {
     pub fn sign(&self) -> Sign {
         super::duration_sign(&self.fields())
     }
+
+    /// DateDurationDays
+    /// TODO(review): Question: what the correct return type?
```
nekevss, line 113: nit: remove the TODO

I believe this was answered in the call that `i64` should be sufficient.

nekevss (approved):
This looks fantastic! Thanks for working on this.

I do have on nit that's not necessarily blocking, and this PR needs to be updated to main, but then this should be good to merge :smile:

src/builtins/core/duration/date.rs
lockels: Yes, i forgot to remove that! thanks reminding me!

PR #187, opened 2025-01-27
Add Eq, Ord impls on FiniteF64

sffc:
Since they are finite, can we implement Eq?

### DIFF
--- a/src/primitive.rs
+++ b/src/primitive.rs
@@ -1,5 +1,7 @@
 //! Implementation of the FiniteF64 primitive
 
+use core::cmp::Ordering;
+
 use crate::{TemporalError, TemporalResult};
 use num_traits::float::FloatCore;
 use num_traits::{AsPrimitive, FromPrimitive, PrimInt};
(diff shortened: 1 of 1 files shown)


src/primitive.rs
```diff
@@ -215,6 +217,20 @@ impl PartialOrd<f64> for FiniteF64 {
...
 }
 
+impl Eq for FiniteF64 {}
+
+impl Ord for FiniteF64 {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.0.partial_cmp(&other.0) {
+            Some(ordering) => ordering,
+            None => {
+                debug_assert!(false, "could not compare fininte f64: {self} {other}");
```
jedel1043, line 227:
```suggestion
                debug_assert!(false, "could not compare finite f64: {self} {other}");
```

jedel1043 (approved):
Looks good! Just a small typo and we can merge this after that.

jasonwilliams (approved):
LGTM, although you will have to remove [partialOrd](https://github.com/boa-dev/temporal/blob/de61400caf1fff5cb97ea489e7eb75b07d8e3a69/src/primitive.rs#L8) from Finite64 (see lint error)

PR #190, opened 2025-02-02
Small API cleanup and a couple dev docs updates

nekevss:
This moves some lingering `as_ixdtf_string` methods to `to_ixdtf_string`. Changes `Instant::from_millisecond` from taking a `i128` to `i64` as that should be more than sufficient for representing an epoch millisecond value in the supported range.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -26,22 +26,18 @@ The primary date & time builtins/components are located in the
 These builtins are then reexported from `lib.rs` to be available from
 `temporal_rs`'s root module.
 
-### Core vs. Native
+### Core vs. Compiled
 
 `temporal_rs`'s builtins are split in two distinct directories `core`
-and `native`. The core implementation contains the core implementation
-of the Temporal builtins; meanwhile, the `native` implementation is a
-Rust wrapper around the `core` implementation that simplifies some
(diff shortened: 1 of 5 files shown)


docs/architecture.md
```diff
@@ -26,23 +26,19 @@ The primary date & time builtins/components are located in the
...
 `temporal_rs`'s builtins are split in two distinct directories `core`
-and `native`. The core implementation contains the core implementation
-of the Temporal builtins; meanwhile, the `native` implementation is a
-Rust wrapper around the `core` implementation that simplifies some
-"lower" level date/time API that may not be necessary for a general use
-case.
+and `compiled`. The core implementation contains the core implementation
+of the Temporal builtins; meanwhile, the `compiled` implementation provides
+method wrappers the `core` methods that simplifies some "lower" level
+date/time API that may not be necessary for a general use case.
```
jedel1043, line 35:
```suggestion
method wrappers around the `core` methods that simplify some "lower" level
date/time APIs that may not be necessary for a general use case.
```

jedel1043 (commented):
Small wording suggestion. Aside from that, looks good!

jedel1043 (approved):

PR #191, opened 2025-02-11
Fix calendar parsing on `from_str` implementation

nekevss:
`parse_allowed_calendar_formats` was returning `Some(None)` instead of `None`. This was preventing calendar ids from correctly moving to `Calendar::from_utf8`

### DIFF
--- a/src/builtins/core/calendar.rs
+++ b/src/builtins/core/calendar.rs
@@ -679,6 +679,7 @@ impl From<PlainYearMonth> for Calendar {
 #[cfg(test)]
 mod tests {
     use crate::{iso::IsoDate, options::TemporalUnit};
+    use core::str::FromStr;
 
     use super::Calendar;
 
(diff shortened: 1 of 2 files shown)


jedel1043:
Uhhh can you remind me again why we have `Option<Option<&[u8]>>`? There are too many empty values there:
- `None`
- `Some(None)`
- `Some(Some(&[]))`

Do we use all of them?

nekevss:
Hmmmm, we're swallowing the error here from the [ParseCalendarString](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporalcalendarstring).

So the return type could be `Result<Option<&[u8]>`, but we are swallowing the error anyways in this instance, so it doesn't offer much utility to return an Error. As it currently stands, the `Some(None)` return was incorrectly signalling a normal completion on the format parsing and going down the wrong code path.

nekevss:
Update: I was able to confirm the `ixdtf` does correctly throw an error on `[u-ca=]`, so the empty value case of `Some(Some(&[]))` is handled during parsing.

Even on the cases that it is not handled, it would be rejected in `Calendar::from_utf8`.

jedel1043:
@nekevss In that case, can we unwrap the `Option`s and use `&[]` as our empty value instead?

nekevss:
> can we flatten the `Option`s and use `&[]` as our empty value instead?

Sure! Actually, I think I like this better. Added a sanity test too for empty calendar annotations.

jedel1043 (approved):
Looks good!

PR #196, opened 2025-02-18
Update time parsing to error on dup critical calendars

nekevss:
Add a time parse variant so that parsing time will throw on critical duplicate calendars.


jedel1043 (approved):
Oh, good catch!

PR #203, opened 2025-02-21
Add `GetNamedTimeZoneTransition` method to TimeZoneProvider trait

nekevss:
This PR adds the stubbed out method to the `TimeZoneProvider` trait, and implements the baseline logic for `ZonedDateTime::get_time_zone_transition`.

This PR also adds the `TransitionDirection` option that lives in the provider module.

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -1,5 +1,7 @@
 //! The `TimeZoneProvider` trait.
 
+use core::str::FromStr;
+
 use crate::{iso::IsoDateTime, time::EpochNanoseconds, TemporalResult};
 use alloc::vec::Vec;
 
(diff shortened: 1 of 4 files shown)


src/provider.rs
```diff
@@ -50,4 +96,13 @@ impl TimeZoneProvider for NeverProvider {
     fn get_named_tz_offset_nanoseconds(&self, _: &str, _: i128) -> TemporalResult<TimeZoneOffset> {
         unimplemented!()
     }
+
+    fn get_named_time_zone_transition(
```
jedel1043, line 100: Should we be consistent on the naming? We use `time_zone` here, but `tz` on the rest of the methods.

nekevss: Yeah, I can change it back for consistency. I'd really like to bikeshed the naming and methods at some point. `tz` really isn't as clear as I'd like it to be, but I'm not sure I'm sold on whether it should be `time_zone`, `timezone` or something else.

raskad (approved):

jedel1043 (approved):

PR #205, opened 2025-02-22
PlainYearMonth parsing and Calendar field resolution cleanup/fixes

nekevss:
This PR fixes some issues with handling `PlainYearMonth`'s calendar field resolution as well as aligning `PlainYearMonth::new_with_overflow` closer to the specification behavior.

This PR does introduce a new internal API `IsoDate::regulate`, which does the regulation work without any date/time limit validations (year month has its own `year_month_within_limits` validation).

Tested on Boa locally, this does increase `Temporal.PlainYearMonth.from` conformance to near 100%. The remaining tests are related to other known issues: minus sign parsing and MonthCode validation.


raskad (approved):
Looks good, good to add the edge case tests

PR #206, opened 2025-02-23
implement utility methods on partial structs

nekevss:
This PR adds methods with the goal of making Partials structs a bit more ergonomic to work with rather than using only struct expressions.

Before:

```rust
let day = Some(24);
let date = PartialDate {
    day,
    ..Default::default()
};
````

After:

```rust
let day = Some(24);
let date = PartialDate::default().with_day(day);
```

If this doesn't make much sense and it's preferred to keep the current method of `StructExpressions`

### DIFF
--- a/src/builtins/core/date.rs
+++ b/src/builtins/core/date.rs
@@ -125,6 +125,44 @@ macro_rules! impl_with_fallback_method {
     };
 }
 
+/// Convenience methods for building a `PartialDate`
+impl PartialDate {
+    pub const fn with_era(mut self, era: Option<TinyAsciiStr<19>>) -> Self {
+        self.era = era;
+        self
+    }
+
+    pub const fn with_era_year(mut self, era_year: Option<i32>) -> Self {
+        self.era_year = era_year;
+        self
+    }
+
+    pub const fn with_year(mut self, year: Option<i32>) -> Self {
+        self.year = year;
+        self
+    }
+
+    pub const fn with_month(mut self, month: Option<u8>) -> Self {
+        self.month = month;
+        self
+    }
+
+    pub const fn with_month_code(mut self, month_code: Option<TinyAsciiStr<4>>) -> Self {
+        self.month_code = month_code;
+        self
+    }
+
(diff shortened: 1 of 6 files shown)


src/builtins/core/duration.rs
```diff
@@ -353,12 +353,6 @@ impl Duration {
         &self.date
     }
 
-    /// Set this `DurationRecord`'s `TimeDuration`.
-    #[inline]
-    pub fn set_time_duration(&mut self, time: TimeDuration) {
-        self.time = time;
-    }
-
```
raskad, line 361: Is this removal intended?

raskad (approved):
I don't mind the addition.

src/builtins/core/duration.rs
nekevss: Removal is intended primarily due to [this comment](https://github.com/boa-dev/temporal/blob/main/temporal_capi/src/duration.rs#L166)

jedel1043 (approved):
I think this makes sense. Moving structs is much safer for FFI than passing references, since it's harder to preserve the mutable borrow XOR shared borrow rule.

src/builtins/core/date.rs
```diff
@@ -125,6 +125,44 @@ macro_rules! impl_with_fallback_method {
...
+        self.day = day;
+        self
+    }
+
+    pub const fn with_calendar(mut self, calendar: Calendar) -> Self {
+        self.calendar = calendar;
+        self
+    }
+}
+
```
jedel1043, line 165: Now that I think about this, can you add a const `new` constructor? Having all of these be `const` is useful, but they're less useful because we don't have a way to ergonomically create a const `PartialDate` without having to manually set all fields to `None`...

jedel1043 (changes requested):

src/builtins/core/date.rs
nekevss: I tried to add a `const` new as well. But it requires a `const` calendar constructor, which was ultimately the blocker. I'll play around with it a bit more.

jedel1043 (approved):
Looks good!

PR #207, opened 2025-02-23
Add `UtcOffset` struct for `PartialZonedDateTime`

nekevss:
This PR introduces to `UtcOffset` parse for use with `PartialZonedDateTime`.

The goal ultimately is providing the tools / API for [PrepareCalendarFields](https://tc39.es/proposal-temporal/#sec-temporal-preparecalendarfields) to be implemented, specifically the ToOffsetString conversion, while also better representing the general invariant of `PartialZonedDateTime`'s offset.

### DIFF
--- a/src/builtins/core/timezone.rs
+++ b/src/builtins/core/timezone.rs
@@ -3,11 +3,13 @@
 use alloc::string::String;
 use alloc::{vec, vec::Vec};
 
+use ixdtf::parsers::records::{TimeZoneRecord, UtcOffsetRecord};
 use num_traits::ToPrimitive;
 
 use crate::builtins::core::duration::DateDuration;
 use crate::parsers::{
-    parse_allowed_timezone_formats, parse_identifier, FormattableOffset, FormattableTime, Precision,
+    parse_allowed_timezone_formats, parse_identifier, parse_offset, FormattableOffset,
+    FormattableTime, Precision,
 };
 use crate::provider::{TimeZoneOffset, TimeZoneProvider};
 use crate::{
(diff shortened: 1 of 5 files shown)


src/builtins/core/timezone.rs
```diff
@@ -21,20 +23,85 @@ use crate::{Calendar, Sign};
...
 
+/// A UTC time zone offset stored in minutes
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct UtcOffset(pub(crate) i16);
+
+impl UtcOffset {
+    pub(crate) fn from_ixdtf_record(record: UtcOffsetRecord) -> Self {
+        // NOTE: ixdtf parser restricts minute/second to 0..=60
+        debug_assert!(record.hour <= 60);
+        debug_assert!(record.minute <= 60);
```
jedel1043, line 34: nit: maybe this should be asserted when creating the `UtcOffsetRecord` instead?

nekevss: This is asserted during the parsing that the values are within a valid range.

The general thought was to use `debug_assert` more throughout the code base in order to assert the invariants of function parameters and hopefully make some debugging easier. But I'm fine with the statements being removed too.

jedel1043: > This is asserted during the parsing that the values are within a valid range.
Then it should be documented as an invariant of `UtcOffsetRecord`. The same way we don't check that a `NonZeroUsize` is zero, we shouldn't check invariants of other types.

jedel1043: I just realized `UtcOffsetRecord` exposes its fields as public, so they could be modified before passing it to this constructor. Should we wrap `UtcOffsetRecord` into an internal type that ensures `minute/second <= 60`? We could also upstream the change, but we shouldn't need to block this until a new version of ixdtf is released.

nekevss: Yeah, the records on `ixdtf` are mostly left `pub`, and `non_exhaustive`.

I'm not entirely convinced that we need to assert this in a new internal type as this constructor is `pub(crate)` the `debug_assert` was mostly to aid in debugging and ensure that the method isn't called incorrectly internally.

jedel1043: Hmmm the problem about adding back the assertions is that we would need to add them on every possible place where `UtcOffsetRecord` is passed as a parameter. Right now is just one usage, but it could be possible that we'll add more usages in the future, and having the assertions contained on a single place would be better from a maintenance perspective.

nekevss: I don't have a strong preference for adding back the `debug_assert`s at current on this PR.

It may be worthwhile adding wrappers in the future that handle the assertions. There's been a need for a more formal parser implementation then currently exists that wraps `ixdtf`'s parser. But that's a bit out of the scope of this PR, imo.

jedel1043: Oh okay, if there's plans for wrapping the parser to formalize things, then we can move invariants there eventually.

jedel1043 (approved):
Looks good!

PR #208, opened 2025-02-23
Integrate `MonthCode` into public API and related adjustments

nekevss:
This change integrate `MonthCode` into the public API of the partial fields and `month_code` methods in order to better represent [PrepareCalendarFields](https://tc39.es/proposal-temporal/#sec-temporal-preparecalendarfields) conversion.

This does also add a host of public APIs to `MonthCode` for interacting with it and adjusts the general validation methods (although, I do suspect this will need to be iterated on in the future to better support more calendars).

CC: @Manishearth as this includes updates to `temporal_capi`


Manishearth (approved):

jedel1043 (approved):
Looks good! I think it's a plus that we're not exposing `TinyStr` directly, so that's good.

PR #209, opened 2025-02-23
Implement the remaining non-ISO calendar method calls

nekevss:
Some of the method calls into `icu_calendar` were not fully implemented / active. This implements the methods that currently can be implemented and marks the methods that either need to be implemented or need more research completed with a TODO.


jedel1043 (approved):
LGTM

PR #212, opened 2025-02-24
Test all combinations of features

jedel1043:
Found a couple of bugs related to feature-gating, so I think testing with all combinations of the features is worth doing.


nekevss (approved):
Oh nice! Good catch!

PR #215, opened 2025-02-25
Cleanup utils module

jedel1043:
This PR introduces an `Epoch` struct on the utils module, which unifies all our APIs into a single utility type.

Marking as draft for now since I'd like to test against Boa first, but it's reviewable.


src/utils.rs
```diff
@@ -1,168 +1,217 @@
...
 
 use crate::MS_PER_DAY;
+pub(crate) const MS_PER_HOUR: i64 = 3_600_000;
+pub(crate) const MS_PER_MINUTE: i64 = 60_000;
 
 mod neri_schneider;
 
-pub(crate) use neri_schneider::epoch_days_from_gregorian_date;
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct Epoch {
```
nekevss, line 16: thought: Potentially rename to `EpochMilliseconds` and upgrade to `pub`.

We have an `EpochNanoseconds` type that's already public, and I was already debating whether to have a public `EpochMillisecond` for `Instant::from_milliseconds` and use `EpochNanoseconds` in `Instant::from_nanoseconds`. This could then result in pulling a majority of the `EpochMilliseconds` type out of `utils` into something like `epoch_primitives`, `epoch`, or something. I'm not really sure about the module name. I'm not in love with where `EpochNanoseconds` is exported from currently either, which I think is `temporal_rs::time`. Heck, `temporal_rs::epoch` probably makes even more sense than `time`.

Doing this then leaves whatever is left in `utils` as pure utility functions.

src/utils.rs
```diff
@@ -1,168 +1,217 @@
-//! Utility date and time equations for Temporal
+#![allow(
+    unused,
+    reason = "prefer to have unused methods instead of having to gate everything behind features"
```
nekevss, line 3: Yeah, probably a better approach.

nekevss (commented):
Really like the work to clean this up. I had a thought when I saw the new `Epoch` struct. Let me know what you think.

src/builtins/core/year_month.rs
```diff
@@ -103,11 +102,19 @@ impl PlainYearMonth {
...
     }
 
-    /// Returns the padded ISO year string
+    /// 3.5.11 PadISOYear ( y )
+    ///
+    /// Returns a String representation of y suitable for inclusion in an ISO 8601 string.
     #[inline]
     #[must_use]
     pub fn padded_iso_year_string(&self) -> String {
-        pad_iso_year(self.iso.year)
```
nekevss, line 110: I think this can actually be removed. We're not using this in `Boa` and `to_ixdtf_string` supersedes it. For some reason, I didn't notice this was still part of the public API.

PR #223, opened 2025-02-28
Fixes for `EpochNanosecond`s and `Offset` parsing

nekevss:
This PR contains some fixes that was causing `IsoDateTime::from_epoch_nanoseconds` to throw incorrectly along with fixing the offset parsing that was being too forgiving and therefore not throwing errors when it should.

### DIFF
--- a/src/parsers/timezone.rs
+++ b/src/parsers/timezone.rs
@@ -77,6 +77,10 @@ pub(crate) fn parse_offset(chars: &mut Peekable<Chars<'_>>) -> TemporalResult<i1
     // First offset portion
     let hours = parse_digit_pair(chars)?;
 
+    if !(0..24).contains(&hours) {
+        return Err(TemporalError::range().with_message("Invalid offset hour value."));
+    }
+
     let sep = chars.peek().is_some_and(|ch| *ch == ':');
     if sep {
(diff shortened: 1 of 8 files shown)


src/parsers/timezone.rs
```diff
@@ -77,6 +77,10 @@ pub(crate) fn parse_offset(chars: &mut Peekable<Chars<'_>>) -> TemporalResult<i1
     // First offset portion
     let hours = parse_digit_pair(chars)?;
 
+    if !(0..24).contains(&hours) {
```
nekevss, line 80: This was implemented in `temporal_rs` because there is not really any exported `OffsetParser` from `ixdtf`. I was hoping this could be pretty minimal, which is greedy. I do think it'd be best to try and upstream this to `ixdtf` if possible, especially to unify the parsing logic. In the meantime, this does fix the bugs related to it.

jedel1043: Can you open an issue for it then? Just to keep track of the things that we need to upstream.

nekevss: Opened #226

jedel1043 (approved):
Thank you!

PR #216, opened 2025-02-25
Add with to PlainYearMonth

sebastianjacmatt:
Worked with @HenrikTennebekk on #143

We encountered an issue in `temporal_capi/src/plain_year_month.rs`.
Error: 
the trait From<options::ffi::ArithmeticOverflow> is not implemented for Option<temporal_rs::options::ArithmeticOverflow>

Unsure how to fix this.

Issue #143, opened 2024-12-25
Implement `with` method for `PlainYearMonth`


nekevss:
This is related to ongoing work for #15

`PlainYearMonth::with` needs to be implemented.

sebastianjacmatt:
Currently working on this issue

PR #231, opened 2025-03-02
Add with to PlainYearMonth

sebastianjacmatt:
Restructured pull request of #216

Worked with @HenrikTennebekk on #143 

Had to change temporal_capi/plain_year_month.rs hope this is fine?

### DIFF
--- a/src/builtins/core/date.rs
+++ b/src/builtins/core/date.rs
@@ -78,6 +78,7 @@ impl PartialDate {
         })
     }
 
+    crate::impl_with_fallback_method!(with_fallback_year_month, PlainYearMonth);
     crate::impl_with_fallback_method!(with_fallback_date, PlainDate);
     crate::impl_with_fallback_method!(with_fallback_datetime, PlainDateTime);
     // TODO: ZonedDateTime
--- a/src/builtins/core/year_month.rs
+++ b/src/builtins/core/year_month.rs
@@ -182,13 +182,35 @@ impl PlainYearMonth {
         self.calendar.identifier()
     }
 
+    /// Returns the calendar day value.
+    pub fn day(&self) -> TemporalResult<u8> {
+        self.calendar.day(&self.iso)
+    }
+
     /// Creates a `PlainYearMonth` using the fields provided from a [`PartialDate`]
     pub fn with(
         &self,
-        _partial: PartialDate,
-        _overflow: ArithmeticOverflow,
+        partial: PartialDate,
+        overflow: Option<ArithmeticOverflow>,
     ) -> TemporalResult<Self> {
-        Err(TemporalError::general("Not yet implemented."))
+        // 1. Let yearMonth be the this value.
+        // 2. Perform ? RequireInternalSlot(yearMonth, [[InitializedTemporalYearMonth]]).
+        // 3. If ? IsPartialTemporalObject(temporalYearMonthLike) is false, throw a TypeError exception.
+        if partial.is_empty() {
+            return Err(TemporalError::r#type().with_message("A PartialDate must have a field."));
+        };
+        // 4. Let calendar be yearMonth.[[Calendar]].
+        // 5. Let fields be ISODateToFields(calendar, yearMonth.[[ISODate]], year-month).
+        // 6. Let partialYearMonth be ? PrepareCalendarFields(calendar, temporalYearMonthLike, « year, month, month-code », « », partial).
+        // 7. Set fields to CalendarMergeFields(calendar, fields, partialYearMonth).
+        // 8. Let resolvedOptions be ? GetOptionsObject(options).
+        // 9. Let overflow be ? GetTemporalOverflowOption(resolvedOptions).
+        // 10. Let isoDate be ? CalendarYearMonthFromFields(calendar, fields, overflow).
+        // 11. Return ! CreateTemporalYearMonth(isoDate, calendar).
+        self.calendar.year_month_from_partial(
+            &partial.with_fallback_year_month(self)?,
+            overflow.unwrap_or(ArithmeticOverflow::Constrain),
+        )
     }
 
     /// Compares one `PlainYearMonth` to another `PlainYearMonth` using their
--- a/src/builtins/core/year_month.rs
+++ b/src/builtins/core/year_month.rs
@@ -322,6 +322,84 @@ mod tests {
 
     use super::PlainYearMonth;
 
+    use tinystr::tinystr;
+
+    use super::*;
+
+    #[test]
+    fn test_plain_year_month_with() {
+        let base = PlainYearMonth::new_with_overflow(
+            2025,
+            3,
+            None,
+            Calendar::default(),
+            ArithmeticOverflow::Reject,
+        )
+        .unwrap();
+
+        // Year
+        let partial = PartialDate {
+            year: Some(2001),
+            ..Default::default()
+        };
+
+        let with_year = base.with(partial, None).unwrap();
+        assert_eq!(with_year.iso_year(), 2001); // year is changed
+        assert_eq!(with_year.iso_month(), 3); // month is not changed
+        assert_eq!(
+            with_year.month_code().unwrap(),
+            MonthCode::from_str("M03").unwrap()
+        ); // assert month code has been initialized correctly
+
+        // Month
+        let partial = PartialDate {
+            month: Some(2),
+            ..Default::default()
+        };
+        let with_month = base.with(partial, None).unwrap();
+        assert_eq!(with_month.iso_year(), 2025); // year is not changed
+        assert_eq!(with_month.iso_month(), 2); // month is changed
+        assert_eq!(
+            with_month.month_code().unwrap(),
+            MonthCode::from_str("M02").unwrap()
+        ); // assert month code has changed as well as month
+
+        // Month Code
+        let partial = PartialDate {
+            month_code: Some(MonthCode(tinystr!(4, "M05"))), // change month to May (5)
+            ..Default::default()
+        };
+        let with_month_code = base.with(partial, None).unwrap();
+        assert_eq!(with_month_code.iso_year(), 2025); // year is not changed
+        assert_eq!(
+            with_month_code.month_code().unwrap(),
+            MonthCode::from_str("M05").unwrap()
+        ); // assert month code has changed
+        assert_eq!(with_month_code.iso_month(), 5); // month is changed as well
+
+        // Day
+        let partial = PartialDate {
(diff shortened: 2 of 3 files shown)


src/builtins/core/year_month.rs
```diff
@@ -300,6 +322,84 @@ mod tests {
 
     use super::PlainYearMonth;
 
+    use tinystr::tinystr;
+
+    use super::*;
+
+    #[test]
+    fn test_plain_year_month_with() {
```
nekevss, line 330: praise: great tests!

src/builtins/core/year_month.rs
```diff
@@ -300,6 +322,84 @@ mod tests {
...
+
+        // Day
+        let partial = PartialDate {
+            day: Some(15),
+            ..Default::default()
+        };
+        let with_day = base.with(partial, None).unwrap();
+        assert_eq!(with_day.iso_year(), 2025); // year is not changed
+        assert_eq!(with_day.iso_month(), 3); // month is not changed
+        assert_eq!(with_day.iso.day, 15); // day is changed
```
nekevss, line 388: issue: so this one is a little tricky, but the ISO day property shouldn't be touched.

For reference, see the [basic PlainYearMonth.prototype.with test](https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/PlainYearMonth/prototype/with/basic.js)

src/builtins/core/date.rs
```diff
@@ -78,6 +78,7 @@ impl PartialDate {
         })
     }
 
+    crate::impl_with_fallback_method!(with_fallback_year_month, PlainYearMonth);
```
nekevss, line 81: issue (part 2 on test review comment): so this macro impl specifically on `PartialDate` is a bit problematic for specifically `PlainYearMonth`.

The method steps at play here are 6-7:
```txt
6. Let partialYearMonth be ? [PrepareCalendarFields](https://tc39.es/proposal-temporal/#sec-temporal-preparecalendarfields)(calendar, temporalYearMonthLike, « year, month, month-code », « », partial).
7. Set fields to [CalendarMergeFields](https://tc39.es/proposal-temporal/#sec-temporal-calendarmergefields)(calendar, fields, partialYearMonth).
```

`PrepareCalendarFields` is a bit of a complex abstract operations that's main purpose is to pull fields from a JavaScript object. In order to handle this, the partial objects were created.

You correctly noted that the partial cannot be empty when submit into `with`, but the actual assertion technically occurs in step 6, not 3.

This method implemented by this macro is `temporal_rs`'s answer for step 7. The only issue is that day should never be set downstream by the caller due to step 6.

The simple solution would be to manually implement a `with_fallback_year_month` that does not touch the `partial.day` (like the other year_month method above this).

The other solution (that after typing this all out I'm more convinced is the correct approach) is to move that invariant of not touching the day field into the type system and create a `PartialYearMonth` struct that can then be used on all `PlainYearMonth` paths that are using `PartialDate`.

The second solution is probably a bit of scope creep, so if you'd like to go with the first solution, that's fine by me. We can review and merge this, and I'll open up another issue for implementing `PartialYearMonth`.

If you want to do the second solution, I think `PartialYearMonth` should look something like:

```
struct PartialYearMonth {
    pub year: Option<i32>,
    pub month: Option<u8>,
    pub month_code: Option<MonthCode>,
    pub era: Option<TinyAsciiStr<19>>,
    pub era_year: Option<i32>,
    pub calendar: Calendar,
}
```

nekevss (changes requested):
This looks fantastic!

I have one issue, but it doesn't really have anything to do with the code, so much as it does some weird oddities with `PlainYearMonth`. I tried to give as detailed of an explanation as I could. Let me know what you'd prefer.

src/builtins/core/year_month.rs
```diff
@@ -182,13 +182,35 @@ impl PlainYearMonth {
         self.calendar.identifier()
     }
 
+    /// Returns the calendar day value.
+    pub fn day(&self) -> TemporalResult<u8> {
```
nekevss, line 186: I was double checking the PR and noticed this method. This should be removed as `PlainYearMonth` has no day accessor on the builtin.

nekevss (commented):
Following up on this PR for the CI failure.

We've started checking in the cpp bindings for `temporal_capi`. Since this PR makes a change to `temporal_capi`. You'll need to update the bindings. To do so, simply run `cargo run -p diplomat-gen`. This will update the bindings and then you just add them to history. Let me know if you have any issues.

src/builtins/core/date.rs
sffc: Another solution could be to make the function return an `Err` if the PartialDate `day` field is populated, and then just use the regular macro?

But I like the type system approach of `PartialYearMonth`

nekevss: > Another solution could be to make the function return an `Err` if the PartialDate `day` field is populated, and then just use the regular macro?

Yeah, the incoming partial would still have to throw on `partail.day.is_some()`, but the regular macro still sets the day according to the fallback, which would still be invalid as well I believe. The day field is "resolved" by the calendar, at least according to the [specification, in step 2](https://tc39.es/proposal-temporal/#sec-temporal-calendaryearmonthfromfields).

src/builtins/core/year_month.rs
sebastianjacmatt: I'm having an issue with the FFI bindings from boa-dev/temporal to boa. When I run `cargo test` in [boa](https://github.com/boa-dev/boa) I get the error:
```rust
core/engine/src/builtins/temporal/plain_year_month/mod.rs:372:53
let result = year_month.inner.with(partial, overflow)?;
expected `Option<ArithmeticOverflow>`, found `ArithmeticOverflow`
```
I ran `cargo run -p diplomat-gen`. PlainYearMonth.hpp shows a change from ArithmeticOverflow to ArithmeticOverflow_option, so it appears the bindings were updated, yet the error persists🤔
Skjermbilde 2025-03-09 kl  13 43 52

nekevss: Ah, that looks like it's the call from `boa_engine` that is causing the compilation error.

You'd need to update the code in Boa. The `get_option` [call](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/temporal/plain_year_month/mod.rs#L368) will need it's `unwrap_or_default` removed

src/builtins/core/date.rs
sebastianjacmatt: I’ve implemented a solution that avoids modifying the `day` field while still using the macro. Instead of implementing a separate `with_fallback_year_month` or type system, I’ve tweaked the macro to require explicit day specification, so PlainYearMonth can use it without setting day. 
It’s set to None via ..Default::default(), but I’m concerned this doesn’t strictly align with spec step 6 of 9.3.13. Is this okay, or should I continue with the type system instead?

Check out the latest commit. I’ve also changed the test cases which now ignores any attempt to change the `day` of a `PlainYearMonth`.

nekevss: I saw the change you had made to the macro :smiley: It's a pretty creative one! 

Strictly aligning to the spec on step 6 is incredibly hard due to its reliance on JavaScript objects, and essentially requires more of an interpretation rather than step by step adherence, which is how [the partial objects came to be](https://github.com/boa-dev/temporal/pull/89).

The best way to test is to see if it passes the test262 test suite. Have you run the test262 test suite with Boa by chance to test conformance?

I still have to do a more proper review, but I think I'm fine with merging this current approach for now. But as mentioned above, I think long term moving to a `PartialYearMonth` struct is the best option overall. So a follow up PR to complete that is definitely best.

sebastianjacmatt: The macro trick was all @sffc—credit goes to him😅. Right now, testing Temporal with Boa doesn’t seem to work for me🤔. I suspect it may relate to the MonthCode change in temporal, I’m not entirely sure. If the fix is trivial, I’ll try and run Test262.

Here are some of the errors i am experiencing

```rust
error[E0277]: the `?` operator can only be used on `Result`s, not `Option`s, in a method that returns `Result`
   --> core/engine/src/builtins/temporal/plain_date/mod.rs:334:19
    |
321 |     fn get_era(this: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
    |     ------------------------------------------------------------------------------- this function returns a `Result`
...
334 |             .era()?
    |                   ^ use `.ok_or(...)?` to provide an error compatible with `Result<value::JsValue, error::JsError>`
    |
    = help: the trait `FromResidual<std::option::Option<Infallible>>` is not implemented for `Result<value::JsValue, error::JsError>`
    = help: the trait `FromResidual<Result<Infallible, E>>` is implemented for `Result<T, F>`

error[E0599]: `TinyAsciiStr<16>` is not an iterator
   --> core/engine/src/builtins/temporal/plain_date/mod.rs:335:14
    |
332 |           Ok(date
    |  ____________-
333 | |             .inner
334 | |             .era()?
335 | |             .map(|s| JsString::from(s.as_str()))
    | |             -^^^ `TinyAsciiStr<16>` is not an iterator
    | |_____________|
    |
    |
   ::: /Users/sebastianmatthews/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tinystr-0.8.1/src/ascii.rs:15:1
    |
15  |   pub struct TinyAsciiStr<const N: usize> {
    |   --------------------------------------- doesn't satisfy `TinyAsciiStr<16>: std::iter::Iterator`
    |
    = note: the following trait bounds were not satisfied:
            `TinyAsciiStr<16>: std::iter::Iterator`
            which is required by `&mut TinyAsciiStr<16>: std::iter::Iterator`
            `str: std::iter::Iterator`
            which is required by `&mut str: std::iter::Iterator`
```

There is a total of 34 errors

nekevss: Has this been rebased or merged with main recently? I thought more or less Boa was up to date ... but maybe its a bit more out of sync then I had thought. I can have an update complete by tonight.

nekevss:
Yeah, I'm not actually sure of a good way to fix this. Although, I am curious if there is a way.

In the meantime, just annotate the macro with the recommend allow clippy lint: `#[allow(clippy::needless_update)]`.

src/builtins/core/date.rs
sebastianjacmatt: I've rebased onto the main branch and successfully tested with the test262 suite.
Note: I modified the `get_option` [call](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/temporal/plain_year_month/mod.rs#L368) in boa, removing the `.unwrap_or_default();`

PR #234, opened 2025-03-04
Add various updates to docs

nekevss:
This updates the docs to include how to run `diplomat-gen`.

It also includes adding an example to the README and the lib.rs header. It may be worthwhile to look into unifying those in the future.

Once we have a good setup for handling `TimeZoneProvider`s, it may be worth while looking into including `Now` into the docs that can run in CI. For now I went with a rather simple example to show conversion of `PlainDate` with an ISO calendar to a `PlainDate` with a japanese calendar.

The examples are nice to see, but they do show a slight weakness in the API ergonomics that using `TemporalResult` is not always ideal. I think the primary reason the `TemporalResult` is there is due to a fallible conversion from us -> ICU4X. But I'm not entirely sure it is 100% fallible. It would be worth confirming if we could use an `expect` on `as_icu4x`. It would probably dramatically increase our API ergonomics.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,3 +22,14 @@ the Temporal specification.
 
 For more information on testing and debugging `temporal_rs`. Please see
 the [testing overview](./docs/testing.md).
+
+## Diplomat and `temporal_capi`
+
+If changes are made to `temporal_capi` that affect the public API, the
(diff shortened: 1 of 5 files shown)


jedel1043 (approved):
Docs look great!

PR #235, opened 2025-03-04
Clean up API ergonomics for calendar methods

nekevss:
This PR removes a lot of the `TemporalResult` returns from most of the calendar methods.

This is made possible on most methods because we no longer need to support Not Yet Implemented errors.

There is also an expect added in the conversion from `IsoDate` -> `icu_calendar::Date<Iso>`. This is tested to be valid in a test added to `iso.rs` for the max bounds of `IsoDate`.

The core changes in this PR are made to `iso.rs` and `calendar.rs` by extension. The rest is mostly updating all the calendar methods accordingly.

Note: calendar methods that are still not fully implemented across various calendars were left as a `TemporalResult`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,7 @@ let iso8601_date = PlainDate::try_new(2025, 3, 3, Calendar::default()).unwrap();
 
 // Create a new date with the japanese calendar
 let japanese_date = iso8601_date.with_calendar(Calendar::from_str("japanese").unwrap()).unwrap();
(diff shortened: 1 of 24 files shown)


jedel1043 (approved):
Good to know we can directly convert to ICU4X's dates without much issue!

nekevss:
Yeah, it's nice to know and have that tested reliably for our max and min dates.

PR #241, opened 2025-03-08
Initial implementation of Duration.prototype.total

### LINKED ISSUE #139: Implement `total` method for `Duration`
Per the [Temporal specification](https://tc39.es/proposal-temporal/#sec-temporal.duration.prototype.total), Duration needs to have a `Duration::total` method implemented.

This is related to ongoing progress for #11

lockels:
Fixes #139 

Tests adapted from sample code in https://tc39.es/proposal-temporal/docs/duration.html under total now pass :)

I have a question in duration.rs about return types!

There's also another question in normalized.rs regarding some code i'd like to refactor that makes this PR quite a bit larger than it needs to be


lockels:
migrated some tests from test262:)

src/builtins/core/duration.rs
```diff
@@ -686,11 +686,82 @@ impl Duration {
...
         &self,
-        _unit: TemporalUnit,
-        _relative_to: Option<RelativeTo>,
-        _provider: &impl TimeZoneProvider,
-    ) -> TemporalResult<i64> {
-        Err(TemporalError::general("Not yet implemented"))
+        unit: TemporalUnit,
+        relative_to: Option<RelativeTo>,
+        provider: &impl TimeZoneProvider,
+        // Review question what is the return type of duration.prototye.total?
```
lockels, line 692: question about return types

src/builtins/core/duration/normalized.rs
```diff
@@ -294,6 +303,18 @@ struct NudgeRecord {
     expanded: bool,
 }
 
+// TODO: Review question:
+// had to create a new nudgerecord that doesn't convert and store to i128 in nudge_calendar_unit(),
```
lockels, line 307: needs refactoring i think

src/builtins/core/duration.rs
```diff
@@ -686,11 +686,82 @@ impl Duration {
...
+        // total seems to return floating point numbers, but the signature was orignally
+        // returning a TemporalResult<i64>?? I changed this, but I'm not sure if this is correct.
+    ) -> TemporalResult<f64> {
```
lockels, line 695: question about return types

src/builtins/core/duration/normalized.rs
```diff
@@ -294,6 +303,18 @@ struct NudgeRecord {
...
+// as well as a duplicate function to nudge_calendar_unit_f64() to make it work :/
+// Any suggestions on how to make it better?
+#[allow(dead_code)]
+struct NudgeRecordF64 {
```
lockels, line 311: my other question, changes related to this needs refactoring i think

src/builtins/core/duration.rs
```diff
@@ -686,11 +686,80 @@ impl Duration {
...
-        _unit: TemporalUnit,
-        _relative_to: Option<RelativeTo>,
-        _provider: &impl TimeZoneProvider,
-    ) -> TemporalResult<i64> {
-        Err(TemporalError::general("Not yet implemented"))
+        unit: TemporalUnit,
+        relative_to: Option<RelativeTo>,
+        provider: &impl TimeZoneProvider,
+        // Review question what is the return type of duration.prototye.total?
+    ) -> TemporalResult<f64> {
```
lockels, line 693: total seems to return floating point numbers according to the temporal documentation, but the signature was orignally returning a TemporalResult<i64>?? I changed this, but I'm not sure if this is correct.

src/builtins/core/duration/normalized.rs
```diff
@@ -294,6 +303,15 @@ struct NudgeRecord {
     expanded: bool,
 }
 
+// TODO: Review question: Code needs to be refactored
+#[allow(dead_code)]
+struct NudgeRecordF64 {
```
lockels, line 308: I had to create a new ``NudgeRecord`` that doesn't convert and store to i128 in nudge_calendar_unit(), as well as a duplicate function to nudge_calendar_unit_f64() to make everything work :/ 

Any suggestions on how to refactor this bit?

src/builtins/core/duration.rs
nekevss: Disclaimer: I haven't fully reviewed the code here, but I'll provide a couple early comments below (so take it with a grain of salt).

Hmm, this is one of the reasons that `Duration` is sometimes hard to reason through and also still retains it's `FiniteF64` fields instead of integer fields.

It's fine to switch off the `i64` return that might have been generous thinking on my part. Although, I would recommend using `FiniteF64` over `f64` for the added invariant checks of `FiniteF64`. If we are going to be exposing floating points in the API long term, I think I'd prefer using that type over `f64`.

My temptation is to remove any reference of `FiniteF64` from these APIs, but after looking through this one. It may need to stay. The biggest question is 100% going to be how to handle 7.5.31 `TotalTimeDuration` in a way that's sane.

src/builtins/core/duration/normalized.rs
nekevss: Hmm, if possible try to adjust the original struct and methods as necessary.

Initial reaction is that it may require taking changing `total: 128` to something like `total: DurationTotal` where the arithmetic of `TotalTimeDuraiton` is completed on the `DurationTotal` type.

General thoughts:

```rust
struct DurationTotal {
    total: 128,
    unit: TemporalUnit,
}

impl DurationTotal {
    pub fn to_total(self) -> FiniteF64 {
        // Create FiniteF64 from fields here, i.e. handle casting here too
    }
}
```

Basically my general thought process is that the more that we can implement without using floating points internally, the better.

src/builtins/core/duration.rs
lockels: Yeah makes sense why'd you'd prefer to use ``FiniteF64`` over ``f64``, I'll modify the code accordingly:)

Regarding ``7.5.31`` would something like this be sufficient? Converts the numerator (``u64``) and denominator (``i128``) to a ``FiniteF64`` first, and then does a safe division on it. There might be a slight bit of loss in precision after performing the arithmetic and casting though 😅

```rust
    pub(crate) fn total(&self, unit: TemporalUnit) -> TemporalResult<FiniteF64> {
        let divisor = unit.as_nanoseconds().temporal_unwrap()?;
        let time_duration = FiniteF64::try_from(self.0)?;
        let divisor = FiniteF64::try_from(divisor)?;
        time_duration.checked_div(&divisor)
    }
```

src/builtins/core/duration/normalized.rs
lockels: hm unsure how that would work with the current implementation, but i'll look into it. I haven't implemented it yet was thinking something more like this:

Returning both the rounded and values for fractional ``total``in a struct and then extract the corresponding values for round and total

```rust
#[derive(Debug)]
struct NudgeRecord {
    normalized: NormalizedDurationRecord,
    _total: Option<DurationTotal>,
    nudge_epoch_ns: i128,
    expanded: bool,
}

#[derive(Debug)]
struct DurationTotal {
    RoundedTotal(i128),
    FractionalTotal(FiniteF64),
}
```

src/builtins/core/duration/normalized.rs
```diff
@@ -197,6 +197,19 @@ impl NormalizedTimeDuration {
         ))
     }
 
+    /// Equivalent: 7.5.31 TotalTimeDuration ( timeDuration, unit )
+    /// TODO Fix: Arithemtic on floating point numbers is not safe. According to NOTE 2 in the spec
+    pub(crate) fn total(&self, unit: TemporalUnit) -> TemporalResult<FiniteF64> {
+        let time_duration = FiniteF64::try_from(self.0)?;
```
nekevss, line 203: thought: Step 2's note basically means that while `try_from` here might not fail. It does have the potential to exceed the safe integer range.

We've already discussed this elsewhere, but now that I'm looking at this. I'd be curious if we could `div_rem` to deal with the total, then a `DurationTotal` could be used:

```rust
/// NOTE: I'm mostly guessing integer types, so it could be changeable
struct DurationTotal {
   quotient: i128, // Q: what is the maximum possible quotient
   remainder: u64,
   unit_nanos: u64
}
```

The fraction can then be reconstructed as a method on `DurationTotal`.

```rust
// pseudo code
impl DurationTotal {
    pub fn to_fractional_total(&self) -> TemporalResult<FiniteF64> {
        let fractional_part = FiniteF64(self.rem) / FiniteF64(self.unit_nanos);
        FiniteF64(self.quotient) + fractional_part
    }
}
```

Let me know if you think I'm missing something. The nice part about this is that we know that rem / unit is within the safe integer range because none of the available units exceed the maximum safe integer, so the operation is safe and we also minimize floating point operations to a specific call ... or we can just return `FinifteF64` that's also a valid option.

Either way, I think the operations listed above may be the better option for the above to avoid any Step 2 issues.

src/builtins/core/duration/normalized.rs
```diff
@@ -289,7 +302,7 @@ impl NormalizedDurationRecord {
 #[derive(Debug)]
 struct NudgeRecord {
     normalized: NormalizedDurationRecord,
-    _total: Option<i128>, // TODO: adjust
+    _total: Option<FiniteF64>,
```
nekevss, line 305: nit: remove "_"

It was underscored because it was not be used. Now that it is in use, we don't need the underscore.

src/builtins/core/duration/normalized.rs
```diff
@@ -946,6 +959,43 @@ impl NormalizedDurationRecord {
...
+                ResolvedRoundingOptions {
+                    largest_unit: unit,
+                    smallest_unit: unit,
+                    increment: RoundingIncrement::default(),
+                    rounding_mode: TemporalRoundingMode::Trunc,
+                },
+            )?;
+
+            // c. Return record.[[Total]].
+            return record._total.temporal_unwrap();
```
nekevss, line 990: question: is total guaranteed to return as a return from `nudge_calendar_unit`?

nekevss (commented):
This is looking great!! I had a couple nits and then another thought on the `DurationTotal`.

src/builtins/core/duration/normalized.rs
lockels: Hey, i've attempted this solution. This part of the spec turned out to a bit more difficult and nuanced than i first assumed! Thanks for pointing it out to me 😅😅

src/builtins/core/duration/normalized.rs
lockels: As far as i can see, there are there are no None in certain branches in ``nudge_calendar_unit``, unlike ``nudge_to_zoned_time``

src/builtins/core/duration/normalized.rs
lockels: > ```rust
>    quotient: i128, // Q: what is the maximum possible quotient
> ```

wouldn't it be the maximum possible value for an i128 divided by the smallest case from ``unit.asnanoseconds()``

so ``2^127 - 1 / 1`` ?

nekevss: > wouldn't it be the maximum possible value for an i128 divided by the smallest case from unit.asnanoseconds()

By definition, yes. But this is more so what are the constraints on the actual size of the inputs. For instance, `Duration` can only be a valid `Duration`. See [`IsValidDuration`](https://tc39.es/proposal-temporal/#sec-isvalidduration) op. So could we get away with truncating that field to a smaller integer based on inputs and add a unit test to verify? I don't have an answer to this btw; this is mostly me thinking out loud via GitHub comment. Probably something to test in the future.

> This part of the spec turned out to a bit more difficult and nuanced than i first assumed!

Yeah :sweat_smile: The handling of `Duration` rounding has taken me the most time to think through and come up with something that I was happy with. I'm still nervous that there might be some edge case that was missed. If you look at the contrib graph for early last year and see it was rather low for a bit, it was probably because I was trying to think through the best way to handle `Duration` rounding and the `NormalizedDuration`/`TimeDuration` part of the spec in a way I was happy with. Also, partially why I was a bit overly opinionated on the PR review.

nekevss (approved):
I think that `DurationTotal` as a concept can potentially be iterated on moving forward, but I think this can be completed in future iterations.

Thanks again for your work on this!

PR #243, opened 2025-03-12
Rename methods on `Now` and add a test

nekevss:
This PR adjusts the `Now` core API and adds a test using mock system data. 

The two goals of this PR are the API and implementing at least one more `Now` test. As far as the API goal, this PR attempts to make the API a bit more explicit and clear. If there's any suggestion for a better method name for the core methods, please feel free to provide feedback.

In the future, we should could add more now unit tests using mock data to ensure the behavior is stable and as expected.

### DIFF
--- a/src/builtins/compiled/now.rs
+++ b/src/builtins/compiled/now.rs
@@ -19,7 +19,7 @@ impl Now {
         let timezone = timezone.unwrap_or(TimeZone::IanaIdentifier(sys::get_system_timezone()?));
         let system_nanos = sys::get_system_nanoseconds()?;
         let epoch_nanos = EpochNanoseconds::try_from(system_nanos)?;
-        Now::plain_datetime_iso_with_provider(epoch_nanos, timezone, &*provider)
+        Now::plain_datetime_iso_with_provider_and_system_info(epoch_nanos, timezone, &*provider)
     }
 
     /// Returns the current system time as a [`PlainDate`] with an optional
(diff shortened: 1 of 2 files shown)


src/builtins/compiled/now.rs
```diff
@@ -47,6 +47,6 @@ impl Now {
         let timezone = timezone.unwrap_or(TimeZone::IanaIdentifier(sys::get_system_timezone()?));
         let system_nanos = sys::get_system_nanoseconds()?;
         let epoch_nanos = EpochNanoseconds::try_from(system_nanos)?;
-        Now::plain_time_iso_with_provider(epoch_nanos, timezone, &*provider)
+        Now::plain_time_iso_with_provider_and_system_info(epoch_nanos, timezone, &*provider)
```
jedel1043, line 50: Thought: I'm not against merging this as it is, but I think it would be worth looking into having a builder API instead of two functions with slightly different arguments. Could be hard the same semantics through FFI though, so it'd require a bit of thinking.

jedel1043 (approved):
Looks good!

src/builtins/compiled/now.rs
nekevss: Yeah, I'm up for adjustments or rethinking `Now`'s API. A builder pattern could actually be really nice. The main goal here was to mostly align the method names a little bit better with the rest of the crate and add tests.

I'm really not in love with the names, but at the very least they are specific. We should open an issue for the builder pattern.

PR #244, opened 2025-03-15
initial implementation of ToZonedDateTime, ToPlainDate, ToPlainTime

### LINKED ISSUE #149: Implement `ToZonedDateTime`, `toPlainDate`, and `toPlainTime` method for `PlainDateTime`
Related to ongoing work for #9.

`PlainDateTime` needs the above `to_x` methods implemented.

lockels:
Fixes #149

### DIFF
--- a/src/builtins/core/datetime.rs
+++ b/src/builtins/core/datetime.rs
@@ -2,18 +2,19 @@
 
 use super::{
     duration::normalized::{NormalizedDurationRecord, NormalizedTimeDuration},
-    Duration, PartialDate, PartialTime, PlainDate, PlainTime,
+    Duration, PartialDate, PartialTime, PlainDate, PlainTime, ZonedDateTime,
 };
 use crate::{
     builtins::core::{calendar::Calendar, Instant},
     iso::{IsoDate, IsoDateTime, IsoTime},
     options::{
-        ArithmeticOverflow, DifferenceOperation, DifferenceSettings, DisplayCalendar,
-        ResolvedRoundingOptions, RoundingOptions, TemporalUnit, ToStringRoundingOptions, UnitGroup,
+        ArithmeticOverflow, DifferenceOperation, DifferenceSettings, Disambiguation,
+        DisplayCalendar, ResolvedRoundingOptions, RoundingOptions, TemporalUnit,
+        ToStringRoundingOptions, UnitGroup,
     },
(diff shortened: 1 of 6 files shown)


src/builtins/core/datetime.rs
```diff
@@ -682,8 +683,33 @@ impl PlainDateTime {
         Ok(Self::new_unchecked(result, self.calendar.clone()))
     }
 
+    pub fn to_zoned_date_time_with_provider(
```
nekevss, line 686: nit: add a `PlainDateTime::to_zoned_date_time` implementation on `PlainDateTime` in compiled.

It looks like I missed adding this stub. So a new module should be added to the compiled directory for `PlainDateTime` and the companion method added.

Worthwhile note: there's no current need to add this method to `temporal_capi` just yet because time zone providers over the FFI is still a work in progress. So only the non provider method implementation is needed.

src/builtins/core/datetime.rs
```diff
@@ -682,8 +683,33 @@ impl PlainDateTime {
...
+        let epoch_ns = time_zone.get_epoch_nanoseconds_for(self.iso, disambiguation, provider)?;
+        // 7. Return ! CreateTemporalZonedDateTime(epochNs, timeZone, dateTime.[[Calendar]]).
+        Ok(ZonedDateTime::new_unchecked(
+            Instant::from(epoch_ns),
+            self.calendar.clone(),
+            time_zone.clone(),
+        ))
+    }
+
+    pub fn to_plain_date(&self) -> TemporalResult<PlainDate> {
```
nekevss, line 702: nit: add a temporal_capi implementation of `to_plain_date` for the FFI

This one may be a little harder because it's dealing with the diplomat bindings, but maybe you can take a crack at it. Changes should be made [here](https://github.com/boa-dev/temporal/blob/main/temporal_capi/src/plain_date_time.rs#L251).

nekevss (commented):
Looks great!

This is about ready to merge, but I do have a couple nits that would be nice to have addressed. Mostly because they could be easily forgotten if not addressed in this PR and it's within scope.

nekevss (approved):
Looks great! Thanks again!

As a heads up, I'm going to open up an issue partially related to this PR about return types used, but I think it's a broader issue across the repository and out of scope of this specific PR.

PR #253, opened 2025-03-23
Add intro documentation to ZonedDateTime and PlainDateTime

nekevss:
This PR adds some intro documentation to `ZonedDateTime` and `PlainDateTime`.

Something I noticed while adding this is that we may benefit from having at least one `TimeZoneProvider` implementation
available by default (this would mean adding `tzdb` to default).


jedel1043 (approved):
Thank you for the docs!

PR #257, opened 2025-04-02
Implement zoneinfo parsing/compilation and add TZif structs

nekevss:
It's time to finally tackle the zoneinfo compilation and tzif sourcing issue. The work here lays the baseline for to handle sourcing. There's still a lot more to add, and maybe some bugs to uncover. But this is also major progress to supporting the ability to source tzifs.

Opening as a draft first for early feedback due to just how dense this actually is. So far, the work here allows us to parse zoneinfo files and compile sets of transition data from the parsed zoneinfo files. There is also related work in provider to setup the tzif structures.

The steps missing from this PR is:
  - [x] All that's left is to bridge the gap from transition sets -> TZIFs,
  - [x] Plug everything into the bakeddata tool or a new connected tool.


zoneinfo/src/tzif.rs
```diff
@@ -0,0 +1,69 @@
...
+use alloc::{collections::btree_set::BTreeSet, vec, vec::Vec};
+use hashbrown::HashSet;
+
+use crate::{types::Transition, SingleLineZone};
+
+/// A version 2 TZif block.
+///
+/// Please note: implementation is very minimal
+#[derive(Debug)]
+pub struct TzifBlockV2 {
```
nekevss, line 17: There are a lot of TZif structs everywhere. I like the convenience of a related struct being in `zoneinfo_compiler`, but I could be convinced that it should live elsewhere.

It might be nice to use structs from `tzif`, but that may require some upstream changes.

zoneinfo/src/zone.rs
```diff
@@ -0,0 +1,506 @@
...
+                offset: lmt_entry.std_offset.as_secs(),
+                dst: false,
+                letter: None,
+                time_type: date.time.time_kind(),
+                format: String::from("LMT"), // This value is constant for the first transition
+            }
+        })
+    }
+
+    pub(crate) fn calculate_transitions_for_year(
```
nekevss, line 152: This method is were a large bulk of the zoneinfo compilation work lives.

I tried my best to document it because it's basically a rough reverse engineering of [`outzone`](https://github.com/eggert/tz/blob/main/zic.c#L3149).

zoneinfo/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub fn to_v2_data_block(&self) -> TzifBlockV2 {
+        if let Some(single_line) = &self.single_line_zone {
+            TzifBlockV2::from_single_line_zone(single_line)
+        } else {
+            TzifBlockV2::from_transition_set(&self.transitions)
+        }
+    }
+}
+
+pub enum ZoneInfoData {
```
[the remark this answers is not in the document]
nekevss: Oh, updated these.

nekevss:
Just to note: I'm working through a couple bugs that I came across when double checking some more TZifs.

zoneinfo/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+            CompiledZone::Transitions(set) => set,
+        };
+
+        let data = tzif::parse_tzif_file(Path::new("/usr/share/zoneinfo/America/Chicago")).unwrap();
+        let fs_transitions = data.data_block2.unwrap().transition_times;
+
+        for (computed, fs) in computed_transitions.iter().zip(fs_transitions.iter()) {
+            assert_eq!(computed.at_time, fs.0);
+        }
+    }
```
jedel1043, line 261: Do we really need to bundle every zoneinfo file? IMO we should just create an artificial sample from the files that covers all the edge cases.

I'm mostly worried that if we test against `/usr/share/zoneinfo`, the tests could break every time a new version of tzdb gets released.

nekevss: I had initially created an artificial zoneinfo file of test cases. My concern after doing so was that if down the line there are changes to any zone used as a test case in the artificial version, then it potentially becomes a very manual process to update the zone versus just dropping the update zoneinfo file into the examples folder and letting git handle the diffing. I was trying to minimize the amount bundled, which is why there are only 4 zoneinfo files included vs. the every file.

On the opposite side of things, creating an artificial file would help to isolate test cases, and probably make things a little less cumbersome to deal with. Comment lines could also then be used to note different edge cases of the test. So I'm definitely open to doing so. I just had some concerns about long term maintenance.

Might have been overthinking on my part though, tbh.

nekevss: Yeah, fair enough. There's a good argument for another tool completely separate from our unit tests to test against all the files.

I'll isolate them into a virtual file and pull the compiled data for the test cases.

nekevss:
I think this is finally ready for merge. Or at least, I'm feeling more confident about the correctness of the data this generates/our path towards correctness.

The full zoneinfo files have been pulled in favor of a custom made zoneinfo file. I do have some serialized tzif data in json files locally that we can check in to test the output against if there's any interest in that approach.

provider/src/tzif.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        for zone_identifier in provider.zone_info.zones.keys() {
+            let _ = zones_set.insert(zone_identifier.clone());
+            identifiers.insert(zone_identifier.clone(), zone_identifier.clone());
+        }
+        for (link, zone) in provider.zone_info.links.iter() {
+            identifiers.insert(link.clone(), zone.clone());
+        }
+
+        let zones: Vec<String> = zones_set.iter().cloned().collect();
```
leftmostcat, line 108: suggestion: It's not necessary to use a set struct and then switch to `Vec` later; since the set is built from unique map keys, its entries are guaranteed to be unique anyhow.

provider/src/tzif.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+#[derive(Debug)]
+pub enum ZoneInfoDataError {
+    Build(ZeroTrieBuildError),
+}
+
+impl ZoneInfoProvider<'_> {
+    pub fn build(tzdata: &Path) -> Result<Self, ZoneInfoDataError> {
+        let mut provider = TzdbDataProvider::try_from_zoneinfo_directory(tzdata).unwrap();
```
leftmostcat, line 94: question: Structurally, it's not really clear why there's a step which constructs the provider, a later step to build data from that provider, and then the raw provider data is reused in later steps. I think I would expect the provider to construct itself wholesale.

zoneinfo/src/lib.rs
```diff
@@ -0,0 +1,505 @@
+// Implementation note: this library is NOT designed to be the most
```
leftmostcat, line 1: suggestion: It would be helpful to amend this comment or to write a doc comment for the module as a whole about why this module is in use over `parse-zoneinfo`. It would serve as a nice architectural record and maybe stop folks like me from asking silly questions.

zoneinfo/src/parser.rs
```diff
@@ -0,0 +1,200 @@
...
+        Self {
+            lines: source.lines().peekable(),
+        }
+    }
+
+    #[allow(clippy::while_let_on_iterator)]
+    pub fn parse(&mut self) -> Result<ZoneInfoCompiler, ZoneInfoParseError> {
+        let mut zoneinfo = ZoneInfoCompiler::default();
+        let mut context = LineParseContext::default();
+        while let Some(line) = self.lines.peek() {
```
leftmostcat, line 157: suggestion: Document why this is `while let` on a `Peekable` instead of a `for line in self.lines`. It took me some work to find where it was used for anything other than `next()` inside the loop. Maybe also move the `allow` directly next to the block.

leftmostcat (changes requested):
For the moment, I'm trying to focus on structural review before dropping any sort of nitpicks in I have a few small suggestions, but one big overall comment: I have a pretty hard time following this code in general. I think a lot of that comes from the contents of tzdb being generally not very easy to understand, but for me that's a pretty good argument for thoroughly document the parsing code.

I'm also unclear on a few design decisions here. Why is the there a finite end date for the range of the built TZifs, since it's in the future and (to my understanding) all current zones are rule-based for current and future dates? I'd at least expect it to be a parameter for the build step so that the TZif conversion code can make (and document) that decision.

Overall, I appreciate that there are some tests here against known-good data, and it seems like correctness isn't an issue! For my part, I'd just like to see more documentation both at the API level and the code level, especially since zoneinfo is not a well-documented or obvious format.

provider/src/tzif.rs
```diff
@@ -0,0 +1,142 @@
...
+#[derive(PartialEq, Debug, Clone, yoke::Yokeable, serde::Serialize, databake::Bake)]
+#[zerovec::skip_derive(Ord)]
+#[zerovec::derive(Debug, Serialize)]
+#[databake(path = temporal_provider::tzif)]
+pub struct ZeroTzif<'data> {
+    pub transitions: ZeroVec<'data, i64>,
+    pub transition_types: ZeroVec<'data, u8>,
+    // NOTE: zoneinfo64 does a fun little bitmap str
+    pub types: ZeroVec<'data, LocalTimeRecord>,
+    pub posix: Cow<'data, str>,
```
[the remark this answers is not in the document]
nekevss: This is a pretty interesting idea. For reference, you might also be interested in #265 then. I think the one concern I may have is duplicating the `PosixTimeZone` structs (from the linked PR) in the provider crate, but that may be more of a structure problem.

provider/src/tzif.rs
nekevss: Hmmmm, yeah, that's a good point.

zoneinfo/src/parser.rs
```diff
@@ -0,0 +1,200 @@
...
+            if line.is_empty() || line.starts_with("#") {
+                // Check if line is empty or a comment
+                //
+                // It is important here that a comment matches on "# "
```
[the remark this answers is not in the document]
nekevss: Initially I had `#PACKRATLIST` as its own check. but there are also other comment lines that seem to have other meanings besides just `#PACKRATLIST` (see the Troll rules in `examples/zoneinfo`). I'm still not entirely sure if the other special case comment lines should be supported or not, so this was a middle ground approach.

zoneinfo/src/types.rs
```diff
@@ -0,0 +1,650 @@
...
+        ctx.exit();
+        value
+    }
+}
+
+impl Default for AbbreviationFormat {
+    fn default() -> Self {
+        Self::String("LMT".to_owned())
+    }
+}
```
leftmostcat, line 218: suggestion: It doesn't appear at a glance that this `Default` impl is used and I'm not sure that `AbbreviationFormat` has a clear default value.

zoneinfo/src/types.rs
```diff
@@ -0,0 +1,650 @@
...
+impl Date {
+    pub fn as_secs(&self) -> i64 {
+        let epoch_days = epoch_days_for_rule_date(self.year, self.month, self.day);
+        utils::epoch_seconds_for_epoch_days(epoch_days)
+    }
+}
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct Time {
+    pub sign: Sign,
```
leftmostcat, line 299: question: The significance of `sign` isn't clear for a type named `Time`. Does this represent a length of time or something else?

zoneinfo/src/types.rs
```diff
@@ -0,0 +1,650 @@
...
+    }
+}
+
+// ==== Rule types ====
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ToYear {
+    Max,
+    Year(u16),
+}
```
leftmostcat, line 377: question: `Option<ToYear>` appears several places. Is there a distinction in behavior between `None` and `Some(ToYear::Max)` in those cases?

zoneinfo/src/tzif.rs
```diff
@@ -0,0 +1,64 @@
...
+            for (index, time_type) in local_time_set.iter().enumerate() {
+                if time_type.offset == transition.offset {
+                    transition_types.push(index as u8);
+                }
+            }
+        }
+
+        let local_time_types = local_time_set
+            .iter()
+            .cloned()
```
leftmostcat, line 48: todo: Use `into_iter()` instead of `iter().cloned()`.

zoneinfo/src/utils.rs
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+        // Assert that y is divisble by 400 to ensure we are returning the correct result.
+        assert_eq!(y % 400, 0);
+        366
+    }
+}
+
+pub(crate) fn in_leap_year(year: i32) -> i32 {
+    mathematical_days_in_year(year) - 365
+}
```
leftmostcat, line 21: suggestion: Rename this or have it return `bool`. The way this is used is a little surprising. Maybe `num_leap_days()`.

zoneinfo/src/parser.rs
```diff
@@ -0,0 +1,201 @@
...
+        ctx: &mut LineParseContext,
+    ) -> Result<T, <T as TryFromStr<LineParseContext>>::Error> {
+        T::try_from_str(self, ctx)
+    }
+}
+
+impl TryFromStr<LineParseContext> for String {
+    type Error = ZoneInfoParseError;
+    fn try_from_str(s: &str, _: &mut LineParseContext) -> Result<Self, Self::Error> {
+        Ok(s.parse::<String>().expect("parse String is infallible"))
```
leftmostcat, line 86: question: Why use `s.parse::<String>().expect(...)` instead of just `String::from(s)`?

leftmostcat (changes requested):
Note: I've tried to use [Conventional Comments](https://conventionalcomments.org/) to differentiate between things that seem reasonable (to me; maybe not to anyone else :slightly_smiling_face:) to address in this PR and things which are broader concerns that might merit discussion and followup. For approval, I'd like to see `question`s answered and `todo`s addressed. `suggestions` and `nitpick`s are things that may be nice to address in this PR, though there's nothing precluding followup for those. I've tried to leave bigger structural things to `thought`s to make it clearer that there's no intent of seeing them addressed immediately.

The biggest concerns for me here are clearer documentation of the purpose and representation of structures and the significance/motivation of intermediate data structs. Overall, it seems like the parser/compiler creates a representation that's not really zoneinfo, not really TZif, and not flexible enough to serve as a generic representation of time zones, and I think it's worth evaluating what purpose it's intended to serve and whether the current structure suits that purpose.

For this PR, mostly I have a few minor tweaks to offer. Nothing that would really make a big difference, just might read a little clearer or save a tiny amount of work at runtime. I really appreciate the tests here; there's a lot of work required to parse all this out and it's really helpful to have some confidence that it's all happening correctly.

provider/src/tzif.rs
nekevss: It's definitely used for `ZoneInfoProvider`, but the intention is by no means to tie the output directly to `ZoneInfoProvider`. That's the core reason for the separate `zoneinfo` crate.

zoneinfo/src/rule.rs
```diff
@@ -0,0 +1,380 @@
...
+                let transition = Transition {
+                    at_time: transition_time,
+                    offset: std_offset.as_secs() + rule.save.as_secs(),
+                    dst: rule.is_dst(),
+                    savings: rule.save,
+                    letter: rule.letter.clone(),
+                    time_type: rule.at.time_kind(),
+                    format: String::new(),
+                };
+                let _ = ordered.insert(transition);
```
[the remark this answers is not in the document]
nekevss: This is mostly to deal with lints for clippy.

provider/src/tzif.rs
```diff
@@ -0,0 +1,137 @@
...
+        // Create a Map of <ZoneId | Link, ZoneId>, this is used later to index
+        for zone_identifier in tzdb_source.compiler.zones.keys() {
+            zones.push(zone_identifier.clone());
+            identifiers.insert(zone_identifier.clone(), zone_identifier.clone());
+        }
+        for (link, zone) in tzdb_source.compiler.links.iter() {
+            identifiers.insert(link.clone(), zone.clone());
+        }
+
+        let identier_map: BTreeMap<Vec<u8>, usize> = identifiers
```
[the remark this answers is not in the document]
nekevss: Hmmmm, I'm a bit confused on the suggestion here. The convert is available for `ZeroAsciiIgnoreCaseTrie<Vec<u8>>`, but I'm not seeing the `FromIterator` impl for `ZeroAsciiIgnoreCasetrie<Store>` in the docs.

provider/src/tzif.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+impl ZeroTzif<'_> {
+    fn from_transition_data(data: &ZoneInfoTransitionData) -> Self {
+        let tzif = data.to_v2_data_block();
+        let transitions = ZeroVec::alloc_from_slice(&tzif.transition_times);
+        let transition_types = ZeroVec::alloc_from_slice(&tzif.transition_types);
+        let mapped_local_records: Vec<LocalTimeRecord> =
+            tzif.local_time_types.iter().map(Into::into).collect();
```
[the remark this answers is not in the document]
nekevss: Yeah, I think one of the really nice things would be to make some updates to [`tzif`](https://crates.io/crates/tzif) such that it could support all of these updates in one shared crate.

The reason I've implemented them here in different places is mostly proof of concept before attempting to upstream them into `tzif`.

zoneinfo/src/lib.rs
```diff
@@ -0,0 +1,559 @@
...
+
+// NOTE: candidate for removal? Should this library offer TZif structs long term?
+//
+// I think I would prefer all of that live in the `tzif` crate, but that will
+// be a process to update. So implement it here, and then upstream it?
+impl ZoneInfoTransitionData {
+    pub fn to_v2_data_block(&self) -> TzifBlockV2 {
+        TzifBlockV2::from_transition_data(self)
+    }
+}
```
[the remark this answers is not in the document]
nekevss: I don't have a real reason to be honest. Recently, I've been going back and forth on `From` implementations vs. explicit method calls.

This could just be a terrible idea on my part.

zoneinfo/src/rule.rs
```diff
@@ -0,0 +1,380 @@
...
+    utils::{self, epoch_seconds_for_epoch_days},
+    zone::ZoneBuildContext,
+};
+
+/// An internal struct for returning the applicable rules found
+/// for a year.
+#[derive(Debug)]
+pub(crate) struct ApplicableRules {
+    // Preloaded saving of the applicable rules' dst
+    pub(crate) saving: Time,
```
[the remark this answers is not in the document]
nekevss: Yeah, "saving" and "savings" are a bit abstract. But after spending so much time looking at this, I'm a bit curious as to whether there's any better name? Maybe offset_adjustment?

zoneinfo/src/types.rs
nekevss: Yeah, I missed this / forgot to remove it when I made changes from earlier versions.

The concept at the time was that first row values were "default" unless specified (not a good approach)

zoneinfo/src/types.rs
nekevss: The `Time` type here is pretty broad and is essentially anything that is `[+|-]hh:mm:ss`, so I guess it depends on the context.

Typically, it's an offset amount or savings amount, while `QualifiedTime` is more of a proper "instant".

zoneinfo/src/types.rs
nekevss: Yep! `max` is a specified value, where `None` is an empty value / "-", which means the `ToYear` is the same as the FROM value.

zoneinfo/src/utils.rs
nekevss: Yeah, this was something pulled over from the Temporal spec :laughing:

zoneinfo/src/parser.rs
nekevss: Good question that I don't have an answer for other than I must have been super tired when I wrote that, to be honest.

zoneinfo/src/lib.rs
```diff
@@ -0,0 +1,559 @@
...
+    #[test]
+    fn test_chicago() {
+        let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+        let mut zoneinfo =
+            ZoneInfoCompiler::from_filepath(manifest_dir.join("examples/zoneinfo")).unwrap();
+
+        // Association is needed.
+        let computed_zoneinfo = zoneinfo.build_zone("America/Chicago");
+
+        let data = tzif::parse_tzif_file(Path::new("/usr/share/zoneinfo/America/Chicago")).unwrap();
```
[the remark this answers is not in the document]
nekevss: I have the transition data generated and stored as json locally for debugging. I'm fine with putting them in this PR if there's a preference for that (I'd personally prefer it).

I was a bit hesitant to throw more data into this specific PR. If the consensus is that it would be better to have the data committed and tested against vs. the reading from the file system, then I can definitely add it.

zoneinfo/src/rule.rs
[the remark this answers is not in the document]
leftmostcat: For what it's worth, I can't reproduce this and I don't see any non-default clippy config in the repo.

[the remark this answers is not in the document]
nekevss: Huh, maybe it's just something configured locally then :confused:

zoneinfo/src/rule.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub fn initialize(rule: Rule) -> Self {
+        Self { rules: vec![rule] }
+    }
+
+    pub fn extend(&mut self, rule: Rule) {
+        self.rules.push(rule);
+    }
+
+    // NOTE: To be precise, we will need the savings value active across year boundaries.
+    pub(crate) fn get_rules_for_year(
```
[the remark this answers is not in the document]
nekevss: So there's two reasons here that lead to this point:

1. The rules have a specific range in their range, so this method is just retrieving the valid rules for a specific year being evaluated.

2. Early on, I began working on this with the misguided approach that it would be nice to calculate the transitions by year for the years in a given arbitrary range. I'm convinced at this point that that was not the best way to go about it, but the general way transitions are calculated still partially work according to that initial approach.

I think there may be a tangible refactor where we loop through zone entries first, determine the year range, then loop through those years, but even with this change, I'm not entirely sure if that would effect how rules are retrieved.

zoneinfo/examples/america-anchorage.json
jedel1043, line 1: Can we remove entries from the transitions table of each example? It would be nice to have something smaller for tests, not just a big file of values that we don't really use.

provider/src/tzif.rs
```diff
@@ -0,0 +1,155 @@
...
+    derive(yoke::Yokeable, serde::Serialize, databake::Bake)
+)]
+#[cfg_attr(feature = "datagen", zerovec::derive(Serialize))]
+#[cfg_attr(feature = "datagen", databake(path = timezone_provider::tzif))]
+pub struct ZeroTzif<'data> {
+    pub transitions: ZeroVec<'data, i64>,
+    pub transition_types: ZeroVec<'data, u8>,
+    // NOTE: zoneinfo64 does a fun little bitmap str
+    pub types: ZeroVec<'data, LocalTimeRecord>,
+    pub posix: &'data str,
```
jedel1043, line 50: I would propose parsing the posix string, instead of leaving it as a raw string. IIRC, zoneinfo64 has the `Rules` field, which could map nicely into a common rules format for tzif and zoneinfo64, right?

provider/src/tzif.rs
```diff
@@ -0,0 +1,155 @@
...
+#[zerovec::make_ule(LocalTimeRecordULE)]
+#[derive(PartialEq, Eq, Debug, Clone, Copy, PartialOrd, Ord)]
+#[cfg_attr(
+    feature = "datagen",
+    derive(yoke::Yokeable, serde::Serialize, databake::Bake)
+)]
+#[cfg_attr(feature = "datagen", databake(path = timezone_provider::tzif))]
+pub struct LocalTimeRecord {
+    pub offset: i64,
+    pub is_dst: bool,
```
jedel1043: #356 should remove the need of having this.

provider/src/tzif.rs
nekevss: I believe this is changed in #265. If it isn't then, it would be easy to add on that follow up. This was primarily a filler at the time.

provider/src/tzif.rs
nekevss: Sounds good. This type is mostly meant to coincide with that until everything has been properly moved around.

zoneinfo/examples/america-anchorage.json
nekevss: I'm not entirely sure ... is there a different testing scheme that would be preferred.

We need to be able to test against the offset, the savings flag (`is_dst`), and the time zone abbreviation for correctness, so every entry in the transitions table is used in the test.

provider/src/tzif.rs
nekevss: > zoneinfo64 has the `Rules` field, which could map nicely into a common rules format for tzif and zoneinfo64, right?

It does have a rules format, but I haven't looked super deeply at it or its purpose. It seemed a little strange or edge case based. The posix struct would map to the `final*` fields in the zone table entry.

EDIT: correction, they have consolidated the rules into a table. So, yes there is potential to map between the two formats if need be.

zoneinfo/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// Extend the current `ZoneInfoCompiler` data from another `ZoneInfoCompiler`.
+    pub fn extend(&mut self, other: Self) {
+        self.rules.extend(other.rules);
+        self.zones.extend(other.zones);
+        self.links.extend(other.links);
+        self.pack_rat.extend(other.pack_rat);
+    }
+}
+
+#[cfg(test)]
```
jedel1043: Probably better to extract this as integration tests instead of internal tests, since they seem to access only public APIs.

zoneinfo/examples/america-anchorage.json
jedel1043: I'm not proposing not testing those things, I'm proposing removing all the entries that are double testing the same things by carefully selecting a smaller range of transitions that has all the things we need to test. This would imply modifying the test data generation code to be a bit more aware about the test cases instead of just dumping everything from `/usr/share/zoneinfo`

zoneinfo/src/lib.rs
nekevss: Yeah, that's true.

zoneinfo/examples/america-anchorage.json
nekevss: The best option to do this would be to manually generate the test data in a slim format, which would contain all the primary transitions in the data, and removes the extra transitions from a bloat format that is normally found in `/usr/share/zoneinfo`.

I'm not exactly clear on how much that may save in the transitions though.

If you're referring to an even smaller subset of transitions from the slim format than it may be possible. Not exactly sure how at this point and it'd probably need to be worked through.

Just to note, while building this and testing this against these test cases, I've hit nearly every primary transition point. It's possible to correct a tricky transition, but in correcting it you've messed up the transition before and after and so on.

EDIT: Updating to slim removed ~6000 lines in https://github.com/boa-dev/temporal/pull/257/commits/1b1ea480341320be6520d68be9826ca9f8c1556a

zoneinfo/tests/test_datagen.rs
```diff
@@ -0,0 +1,218 @@
...
+
+// Uncomment and adjust path to generate any new testing data
+//
+// #[test]
+// #[cfg(feature = "std")]
+// fn gen() {
+//     let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+//     let path = manifest_dir.join("tzdata/test/");
+//     generate_test_data(path, "Antarctica/Troll");
+// }
```
jedel1043: This should probably be an external tool like `diplomat-gen`, more so because we could automatically check that the test data is synced with the tzif data.

nekevss: Not against making it an external tool, but it is currently a bit flawed. Probably a better way to manage it.

> we could automatically check that the test data is synced with the tzif data.

What tzif data is being referred to here as far as syncing?

jedel1043: The tzif data that is used to generate the JSON test data

nekevss: Yeah, sourcing that tzif data is going to be the big question there though :laughing:

jedel1043: Is the `zic` binary tied to the tzdb version? Because we could require that you install that tool, then the script would download a pinned version of tzdb to compile it to a tzif

nekevss: `zic` is not pinned to the version, but it has various compilation flags. For instance, currently the files we have in the repo are bloat=SLIM and vanguard. I'm on an Ubuntu distro and that's bloat=FAT and rearguard.

We could download the zoneinfo files, but then that probably should be behind a `networking` feature at some point, which is why the data is precompiled here in the repo.

EDIT: actually zic being not pinned to the version is technically 100% true. They both live in the same repo, so  technically it can be tied.

jedel1043: > We could download the zoneinfo files, but then that probably should be behind a networking feature at some point, which is why the data is precompiled here in the repo.

I don't know if we need a flag if the tool will only be used internally to generate test data... Though, if we decide to expose the tool, I agree that it would be better to have a `networking` feature.

jedel1043: Ohh, I did say an "external tool"... Whoops, I was talking about an internal repo tool that is external to the zoneinfo crate

nekevss: Yeah, I understood what you meant. I think the networking thing is just general hesitation on my part leaking in :laughing: 

I should be able to pull it out later tonight. It would be nicer to work with as some tooling vs. running test (I've just been enjoying clicking the VS Code "Run test" button).

nekevss: Moved it out and added a little bit to a README file for the tool.

PR #258, opened 2025-04-02
Implementation of toZonedDateTimeISO for Instant

### LINKED ISSUE #246: Implement `toZonedDateTimeISO` for `Instant`
The method can be found [here](https://github.com/boa-dev/temporal/blob/main/src/builtins/core/instant.rs#L239). 

It should be implemented according to the [specification](https://tc39.es/proposal-temporal/#sec-temporal.instant.prototype.tozoneddatetimeiso).

The method may need to be adjusted to be a `with_provider` API.

lockels:
Fixes #246 

Implementation of toZonedDateTimeISO for Instant.

most tests pass, one strange test262 function looks like it should pass, but doesn't somehow.

https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Instant/prototype/toZonedDateTimeISO/timezone-string-datetime.js

This one in particular.. The code does indeed throw a range error when trying to create a timezone from those strings.. but the test doesn't pass.

### DIFF
--- a/src/builtins/core/instant.rs
+++ b/src/builtins/core/instant.rs
@@ -237,7 +237,6 @@ impl Instant {
 
     // TODO: May end up needing a provider API during impl
     pub fn to_zoned_date_time_iso(&self, time_zone: TimeZone) -> TemporalResult<ZonedDateTime> {
-        // 4. Return ! CreateTemporalZonedDateTime(instant.[[EpochNanoseconds]], timeZone, "iso8601").
         Ok(ZonedDateTime::new_unchecked(
             *self,
             Calendar::from_utf8(b"iso8601")?,
--- a/src/builtins/core/instant.rs
+++ b/src/builtins/core/instant.rs
@@ -341,7 +341,7 @@ mod tests {
         options::{DifferenceSettings, TemporalRoundingMode, TemporalUnit},
         primitive::FiniteF64,
         time::EpochNanoseconds,
-        TimeZone, NS_MAX_INSTANT, NS_MIN_INSTANT,
+        NS_MAX_INSTANT, NS_MIN_INSTANT,
     };
 
     #[test]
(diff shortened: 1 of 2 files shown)


lockels:
Huh.. strange that one of the tests fail here..

nekevss:
Huh, are you running on an ubuntu distro? Also, have you rebased to main for this PR?

nekevss:
> most tests pass, one strange test262 function looks like it should pass, but doesn't somehow.

There are some parsing bugs for time zones that should be fixed either when we update `ixdtf` or the work on time zone normalization is completed. See #251 and #257 if you're curious.

src/builtins/core/timezone.rs
```diff
@@ -105,7 +105,8 @@ impl TimeZone {
     }
 
     pub fn try_from_str(src: &str) -> TemporalResult<Self> {
-        if let Ok(timezone) = Self::try_from_identifier_str(src) {
+        let normalized = src.to_uppercase();
```
nekevss, line 108: Oh, wait this is probably what's causing the failure. 

So this is actually sort of a tricky case.

If you upper case a time zone string, "America/New_York" becomes "AMERICA/NEW_YORK". The file system will throw on this because it's case sensitive (although, this may actually depend on your OS configuration). That's actually what #251 is working towards, providing a case insensitive way to return the normalized time zone strings.

lockels:
> Huh, are you running on an ubuntu distro?

Nope, makes sense why it fails since i'm on a mac. 

> Also, have you rebased to main for this PR?

Nope, but i think i managed to fix that now.

I'll convert this PR to a draft for now..

src/builtins/core/instant.rs
```diff
@@ -236,8 +236,12 @@ impl Instant {
     }
 
     // TODO: May end up needing a provider API during impl
-    pub fn to_zoned_date_time_iso(&self, _time_zone: TimeZone) -> TemporalResult<ZonedDateTime> {
-        Err(TemporalError::general("Not yet implemented"))
+    pub fn to_zoned_date_time_iso(&self, time_zone: TimeZone) -> TemporalResult<ZonedDateTime> {
```
nekevss, line 239: thought: I think this can be made non-fallible.

`Calendar::from_utf8(b"iso8601")` can be `Calendar::default()`, and then this method returns `ZonedDateTime` directly without a result.

This is inline with the specification:
```
8.3.15 Temporal.Instant.prototype.toZonedDateTimeISO ( timeZone )

This method performs the following steps when called:

    1. Let instant be the this value.
    2. Perform ? RequireInternalSlot(instant, [[InitializedTemporalInstant]]).
    3. Set timeZone to ? ToTemporalTimeZoneIdentifier(timeZone).
    4. Return ! CreateTemporalZonedDateTime(instant.[[EpochNanoseconds]], timeZone, "iso8601").
```

Note how the major step we are handling here is step 4, which is non-fallible. The fallible portion is in step 2 (engine specific step) and step 3 (handled by the conversion to `TimeZone`).

nekevss (approved):
Looks great! Thanks as always!

PR #259, opened 2025-04-08
Since until

### LINKED ISSUE #142: Implement `since` and `until` methods for `PlainYearMonth`
This is related to ongoing work for #15.

`PlainYearMonth::since` and `PlainYearMonth::until` need to be implemented along with any related abstract operations.

Each of these tests account for 72 test262 tests.

sebastianjacmatt:
fixes #142 

Worked with @HenrikTennebekk and @Magnus-Fjeldstad 
Implementation for since and until in PlainYearMonth, has not yet been tested with boa test262


src/builtins/core/year_month.rs
```diff
@@ -59,12 +65,72 @@ impl PlainYearMonth {
...
+        // 12. Let otherDate be ? CalendarDateFromFields(calendar, otherFields, constrain).
+        // 13. Let dateDifference be CalendarDateUntil(calendar, thisDate, otherDate, settings.[[LargestUnit]]).
+        // 14. Let yearsMonthsDifference be ! AdjustDateDurationRecord(dateDifference, 0, 0).
+        let result = self
+            .calendar()
+            .date_until(&self.iso, &other.iso, resolved.largest_unit)?;
+        // 15. Let duration be CombineDateAndTimeDuration(yearsMonthsDifference, 0).
+        let mut duration = NormalizedDurationRecord::from_date_duration(*result.date())?;
+        // 16. If settings.[[SmallestUnit]] is not month or settings.[[RoundingIncrement]] ≠ 1, then
+        if settings.smallest_unit != Some(TemporalUnit::Month)
```
nekevss, line 109: issue: I believe this should be using the resolved options in `resolved`.

src/options.rs
```diff
@@ -178,11 +178,25 @@ impl ResolvedRoundingOptions {
...
         operation: DifferenceOperation,
         unit_group: UnitGroup,
+        dissallow_week_and_day: bool,
         fallback_largest: TemporalUnit,
         fallback_smallest: TemporalUnit,
     ) -> TemporalResult<Self> {
-        // 1. NOTE: The following steps read options and perform independent validation in alphabetical order.
-        // 2. Let largestUnit be ? GetTemporalUnitValuedOption(options, "largestUnit", unitGroup, auto).
+        // Week and day is the only use case where any of the units are not allowed.
+        if dissallow_week_and_day {
```
nekevss, line 186: thought: is there anything preventing moving this into `PlainYearMonth::diff` vs. where it's currently at in `ResolvedRoundingOptions::from_diff_settings`.

This only affects the one code path from `PlainYearMonth`, I think, so maybe this could work in `PlainYearMonth::diff`. This would then avoid the check occurring for the other callers.

nekevss (changes requested):
Looks great! All the extra tests looks great!

I had one thing I noticed and then a general thought/nit (this one isn't necessarily blocking though). Although, it does look like there are some clippy lints that need to be fixed as well to be able to merge.

nekevss (approved):
Looks good to me! Thanks for working on this!

PR #262, opened 2025-04-18
Add `IsoDate::try_balance` method 

### LINKED ISSUE #250: Bug in `IsoDate::add_date_duration` on max range calculations
Currently, `Duration::total` and `Duration::total_with_provider` panics when adding a duration that would exceed the date time range.

## Example case:

```rust
// relativeto-plaindate-add24hourdaystonormalizedtimeduration-out-of-range.js
#[test]
fn add_normalized_time_duration_out_of_range() {
    let duration = Duration::from_partial_duration(PartialDuration {
        years: Some(FiniteF64::from(1)),
        seconds: Some(FiniteF64::try_from(9_007_199_254_740_990_i64).unwrap()),
        ..Default::default()
    }).unwrap();

    let relative_to = PlainDate::new(2000, 1, 1, Calendar::default()).unwrap();

    let err = duration.total(TemporalUnit::Day, Some(RelativeTo::PlainDate(relative_to.clone())));
    assert!(err.is_err())
}

```

## Current behavior: 

The thread panics (in debug)

## Expected behavior:

A range error should be returned by `Duration::total`

nekevss:
This PR closes #250. By fixing the current panicking behavior of the test case.

For background, the current specification lists `BalanceISODate` as fault tolerant. With the neri-schneider calculations, we ultimately don't support calculating a date with epoch days of 2^53 - 1. So instead of waiting for 12.2.6 CalendarDateAdd Step 3 to throw invalid epoch days (1,000,000,001 epoch days), this adds a new `try_balance` method to throw invalid epoch days early when calculating the epoch days.

To note, I have a feeling that there is most likely more changes to be made along this general line related to `day` being represented by an `i32` vs `i64`, but I believe those would be linked to changes related to #189.

### DIFF
--- a/src/builtins/compiled/duration/tests.rs
+++ b/src/builtins/compiled/duration/tests.rs
@@ -787,3 +787,19 @@ fn balance_days_up_to_both_years_and_months() {
         -2.0
     );
 }
+
+// relativeto-plaindate-add24hourdaystonormalizedtimeduration-out-of-range.js
+#[test]
+fn add_normalized_time_duration_out_of_range() {
+    let duration = Duration::from_partial_duration(PartialDuration {
+        years: Some(FiniteF64::from(1)),
+        seconds: Some(FiniteF64::try_from(9_007_199_254_740_990_i64).unwrap()),
+        ..Default::default()
+    })
+    .unwrap();
+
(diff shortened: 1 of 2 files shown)


nekevss:
Closing this in favor of #268 as it addresses this concern and then some.

PR #263, opened 2025-04-18
Add ISO specific constructors to builtins

### LINKED ISSUE #252: Implement ISO specific constructors `new_iso` and `try_new_iso`
It would be nice from a Rust API perspective to offer ISO specific constructors. All our constructors currently require the user to explicitly provide a calendar. Eventually, there may be users that don't particularly care about the calendar.

```rust
// Before
let date = PlainDate::try_new(2025, 3, 23, Calendar::default()).unwrap();

// Proposed
let date = PlainDate::try_new_iso(2025, 3, 23).unwrap();
```

This could drastically simplifies the API for a general purpose use case.

nekevss:
This PR adds new ISO specific constructors for the built ins. It would close #252.


src/builtins/core/datetime.rs
```diff
@@ -277,7 +278,36 @@ impl PlainDateTime {
         )
     }
 
+    /// Creates a new `DateTime` with an ISO 8601 calendar, constraining any arguments that into a valid range.
```
jedel1043, line 281: Saw a couple of these throughout the PR, so just commenting on this one.

```suggestion
    /// Creates a new `DateTime` with an ISO 8601 calendar, constraining any arguments that may be invalid into a valid range.
```

jedel1043 (commented):
Looks good, I just have a suggested fix for some of the docs.

jedel1043 (approved):
Looks great!

PR #264, opened 2025-04-19
Generate the baked zone info data for ZoneInfoProvider

nekevss:
This PR is the follow up for `impl-zoneinfo-support` once merged with the actual baked data generated with the bakeddata tool.


nekevss:
Just to note, it may be better to merge #265 first and update the data in this PR.

jasonwilliams:
I think it would be good to describe exactly how you generated this with some steps, or even update the documentation, in case someone else needs to re-do it.

nekevss:
I think it's going to be best to merge #265 prior to merging the data build

nekevss:
This is finally updated and ready for review. This polishes the general functionality for generating compiled tzdb data and generates the data.

The major changes are to the Rust code. The debug data is generated by `bakeddata` and attempts to make the general review process a bit more sane.

PR #265, opened 2025-04-20
Add POSIX time zone string support to zoneinfo

nekevss:
This PR adds support for POSIX time zone strings based off the [GNU tz variable documentation](https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html).


nekevss:
This is finally updated and more or less ready for review.

I ended up going with a struct for the `provider`'s `ZeroTzif`. We may need to iterate and improve on it in the future, but I do think it's a good baseline.

zoneinfo/src/types.rs
```diff
@@ -205,9 +205,53 @@ pub enum Sign {
...
+    pub(crate) const fn one() -> Self {
+        Time {
+            sign: Sign::Positive,
+            hour: 1,
+            minute: 0,
+            second: 0,
+        }
+    }
+
+    pub(crate) const fn two() -> Self {
```
HalidOdat: Nitpick:
```suggestion
    pub(crate) const fn one_hour() -> Self {
        Time {
            sign: Sign::Positive,
            hour: 1,
            minute: 0,
            second: 0,
        }
    }

    pub(crate) const fn two_hour() -> Self {
```

zoneinfo/src/posix.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub start: PosixDateTime,
+    /// The end time for the transition
+    pub end: PosixDateTime,
+}
+
+#[non_exhaustive]
+#[derive(Debug, PartialEq, Clone)]
+pub struct PosixAbbreviation {
+    /// Flag whether formatted abbreviation is numeric
+    pub is_numeric: bool,
```
HalidOdat, line 124: question: Sorry if this is a dump question... is numerical posix abbreviation defined somewhere else, I looked into the link, couldn't find the syntax (`<...>`)

zoneinfo/src/posix.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl PosixDate {
+    pub(crate) fn from_rule(rule: &Rule) -> Self {
+        match rule.on_date {
+            DayOfMonth::Day(day) if rule.in_month == Month::Jan || rule.in_month == Month::Feb => {
+                PosixDate::JulianNoLeap(month_to_day(rule.in_month as u8, 1) as u16 + day as u16)
+            }
+            DayOfMonth::Day(day) => {
+                PosixDate::JulianNoLeap(month_to_day(rule.in_month as u8, 1) as u16 + day as u16)
```
HalidOdat: These branches are the same... Is this intended? Shouldn't one of these be a `PosixDate::JulianLeap`?

```suggestion
                PosixDate::JulianNoLeap(month_to_day(rule.in_month as u8, 1) as u16 + day as u16)
            }
            DayOfMonth::Day(day) => {
                PosixDate::JulianNoLeap(month_to_day(rule.in_month as u8, 1) as u16 + day as u16)
```

HalidOdat (approved):
Looks good to me!

Just had a minor nitpick and some questions :smile:

zoneinfo/src/posix.rs
nekevss: Ah, true.

That's outlined a little bit more in IANA's theory.html

https://data.iana.org/time-zones/theory.html

zoneinfo/src/posix.rs
nekevss: :scream: 

EDIT: Nice catch!

zoneinfo/src/posix.rs
nekevss: I should also note, there may be more on it somewhere else in the GNU docs, but the one linked in the PR description and the theory files are probably the best sources.

PR #268, opened 2025-04-25
Update Duration's inner representation from floating point to integers.

### LINKED ISSUE #250: Bug in `IsoDate::add_date_duration` on max range calculations

nekevss:
This begins to address #189 and closes #250.

This completely updates the inner representation for `Duration`, migrating from `FiniteF64`s to `i64`. Due to the change there were also some fairly robust changes across the internals for `Duration`.

There were also some related updates made to some parts of rounding to bring it more in line with the specification and also calculate values without needing fractional values.

### DIFF
--- a/src/builtins/core/date.rs
+++ b/src/builtins/core/date.rs
@@ -231,9 +231,12 @@ impl PlainDate {
         //    duration.[[Nanoseconds]]).
         // 6. Let days be duration.[[Days]] + BalanceTimeDuration(norm,
         //    "day").[[Days]].
-        let days = duration.days().saturating_add(
-            TimeDuration::from_normalized(duration.time().to_normalized(), Unit::Day)?.0,
-        );
+        let days = duration
+            .days()
+            .checked_add(
+                TimeDuration::from_normalized(duration.time().to_normalized(), Unit::Day)?.0,
+            )
+            .ok_or(TemporalError::range())?;
 
         // 7. Let result be ? AddISODate(plainDate.[[ISOYear]], plainDate.[[ISOMonth]], plainDate.[[ISODay]], 0, 0, 0, days, overflow).
         let result = self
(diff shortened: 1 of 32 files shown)


src/builtins/core/date.rs
```diff
@@ -235,20 +231,14 @@ impl PlainDate {
         //    duration.[[Nanoseconds]]).
         // 6. Let days be duration.[[Days]] + BalanceTimeDuration(norm,
         //    "day").[[Days]].
-        let days = duration.days().checked_add(
-            &TimeDuration::from_normalized(duration.time().to_normalized(), Unit::Day)?.0,
-        )?;
+        let days = duration.days().saturating_add(
```
nekevss, line 234: `saturating_add` actually has me a little concerned looking back as it may hide some nasty bugs in the future. I'm going to add a method to address this.

jasonwilliams (approved):
LGTM

nekevss:
I'm not going to merge this for now. I was doing some testing locally with Boa, and there's a couple more edge cases that I need to look into.

PR #269, opened 2025-04-30
Expose equals and compare over FFI

Magnus-Fjeldstad:
Solves #266 

Is there a way to add tests for these new method? We belive we found a solution for the issue, but we havent been able to test it. We would like some feedback.

Worked with @HenrikTennebekk


sffc:
Thanks for regenerating the FFI bindings

Probably good to add C++ tests in https://github.com/boa-dev/temporal/blob/main/temporal_capi/cpp_tests/simple.cpp

CC @Manishearth

temporal_capi/src/plain_date.rs
```diff
@@ -155,6 +155,10 @@ pub mod ffi {
                 .map_err(Into::into)
         }
 
+        pub fn equals(&self, other: &Self) -> bool {
```
Manishearth, line 159: thought: I don't actually think these methods are necessary, but I guess it doesn't harm to have them since we're trying to match the API

temporal_capi/src/plain_year_month.rs
```diff
@@ -142,6 +142,28 @@ pub mod ffi {
...
+        pub fn compare(one: &Self, two: &Self) -> i32 {
+            Self::compare_iso_year_month(
+                one.iso_year(),
+                one.iso_month(),
+                two.iso_year(),
+                two.iso_month(),
+            )
+        }
+
+        pub fn compare_iso_year_month(year1: i32, month1: u8, year2: i32, month2: u8) -> i32 {
```
Manishearth, line 154: suggestion: write this as a match on `(year1, month1).cmp((year2, month2))` instead

And link to the spec entry

Manishearth: also, thought: I don't think it would be a bad idea to just have a method return std::cmp::Ordering here, that converts to the integer anyway, and makes it easier to add comparator support later

temporal_capi/src/plain_date_time.rs
```diff
@@ -242,6 +242,74 @@ pub mod ffi {
                 .map_err(Into::into)
         }
 
+        pub fn equals(&self, other: &Self) -> bool {
+            self.0 == other.0
+        }
+
+        pub fn compare(one: &Self, two: &Self) -> i32 {
```
nekevss, line 249: suggestion: I believe there should be a simpler way to implement these.

The built-ins that have an `IsoDate` internally should have a method called `compare_iso` that does this work for you.

The exception to this I believe is the time oriented types, which should implement `PartialOrd` and `Ord`. In these cases, you can use `Ord`'s method `cmp` and cast `Ordering` into an `i8`. Although, `ZonedDateTime` implements a `compare_instant` method instead of an `Ord` implementation.

temporal_capi/src/plain_month_day.rs
```diff
@@ -44,6 +44,33 @@ pub mod ffi {
...
+                1
+            } else if month1 < month2 {
+                -1
+            } else if day1 > day2 {
+                1
+            } else if day1 < day2 {
+                -1
+            } else {
+                0
+            }
```
sffc, line 72: Issue: PlainMonthDay does not have a compare function

temporal_capi/src/plain_date.rs
```diff
@@ -155,6 +156,39 @@ pub mod ffi {
...
+                one.iso_year(),
+                one.iso_month(),
+                one.iso_day(),
+                two.iso_year(),
+                two.iso_month(),
+                two.iso_day(),
+            )
+        }
+
+        pub fn compare_iso_date(
```
sffc, line 174: Please use the library's compare function. Don't write your own. only write a thing wrapper.

temporal_capi/src/plain_date_time.rs
```diff
@@ -242,6 +242,74 @@ pub mod ffi {
...
+                second1.cmp(&second2),
+                millisecond1.cmp(&millisecond2),
+                microsecond1.cmp(&microsecond2),
+                nanosecond1.cmp(&nanosecond2),
+            ];
+            comparisons
+                .iter()
+                .find(|&&ord| ord != std::cmp::Ordering::Equal)
+                .map_or(0, |ord| match ord {
+                    std::cmp::Ordering::Greater => 1,
```
nekevss, line 307: thought: `Ordering` is `repr(i8)`, so you should be able to cast it. Somthing like: `i32::from(ord as i8)`

nekevss (commented):
Thanks for working on this! I'm adding a couple early review comments as I believe you're still looking at adding some tests.

I think using the native rust methods wherever possible is the best approach as it assures that the compare methods work the same between the Rust implementation and the FFI.

EDIT: Just noticed, Manish and Shane's comments. Feel free to disregard my review coments as I believe we have the same comments.

temporal_capi/src/plain_date_time.rs
Manishearth: Also `diplomat` supports returning `Ordering` over FFI, since it supports custom comparators.

nekevss: Ooooooooh, good to know!

Magnus-Fjeldstad:
Tried to follow your feedback, please give me more feedback if i havent implemented it correctly.

@nekevss @Manishearth

temporal_capi/src/plain_month_day.rs
```diff
@@ -44,6 +44,26 @@ pub mod ffi {
...
+                one.iso_year(),
+                one.iso_month(),
+                two.iso_year(),
+                two.iso_month(),
+            )
+        }
+
+        pub fn compare_iso_year_month(year1: i32, month1: u8, year2: i32, month2: u8) -> i32 {
+            match (year1, month1).cmp(&(year2, month2)) {
+                std::cmp::Ordering::Less => -1,
```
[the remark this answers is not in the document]
Magnus-Fjeldstad: Hey, thanks for the respone, should i do this for all the files.
I.e: 
plain_date_time.rs
plain_date.rs
plain_month_day.rs
plain_time.rs
plain_year_month.rs

im sorry if this is a might be a silly question, but im not so used to the github lingo just yet

Here is my proposed new code for compare 

```rust
pub fn compare_iso_year_month(
            year1: i32,
            month1: u8,
            year2: i32,
            month2: u8,
        ) -> core::cmp::Ordering {
            (year1, month1).cmp(&(year2, month2))
        }

        pub fn compare(one: &Self, two: &Self) -> core::cmp::Ordering {
            Self::compare_iso_year_month(
                one.iso_year(),
                one.iso_month(),
                two.iso_year(),
                two.iso_month(),
            )
        }
        ```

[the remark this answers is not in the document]
Manishearth: Yes. Is there a reason you want separate compare vs compare_iso_year_month methods? It doesn't seem like the spec needs them, just `compare()` is fine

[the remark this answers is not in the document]
nekevss: For reference why `compare_iso` vs. `compare`.

https://github.com/boa-dev/temporal/pull/175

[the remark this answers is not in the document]
Manishearth: I don't mind the name, it's specifically that we have two functions. We should just have one.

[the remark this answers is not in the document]
Magnus-Fjeldstad: Made them into one function now, also used `core` insted of `std`

Manishearth (approved):

nekevss (approved):

nekevss:
This just needs a rebase / merge to resolve the conflicts, and then it should be good to go.

PR #292, opened 2025-05-09
Bump to icu_calendar 2.0

nekevss:
This PR bumps `icu_calendar` to 2.0.

There are a host of changes due to updates in `icu_calendar`.

The ICU calendar trait is now sealed, so we can not implement it anymore. Although, I'm not sure we need to anymore.

The way we handle `from_str` and `from_utf8` are also updated as well to use `CalendarAlgorithm`. This aligns better with our expected values, but it also means we are pulling in `icu_locale` as well.


nekevss:
Forgot to update `temporal_capi`. Will updated ASAP

jedel1043 (approved):
Nice!

PR #296, opened 2025-05-10
Implement a builder API for Now

### LINKED ISSUE #249: Explore a builder pattern API for `Now`
Per review in #243.

`Now` currently has a very verbose and overly explicit API. It could be useful exploring a builder pattern for `Now` to improve the API.

nekevss:
This PR implements a builder API for now and closes #249.

It has some notable changes to the approach to `Now`.

Mainly `Now` is essentially cannot be built without using the new `NowBuilder` and any method called on `Now` consumes it.

For general purpose, a `Temporal` namespace struct was created with `now` and `try_now` methods, which uses the `NowBuilder` to construct a `Now` with values from the system.

This does differ a bit from the specification in that we are eagerly calling the host system time zone, rather than lazily when the user time zone is `None`. But I think this can still be worked around on the engine side if lazily calling is preferred and is more a place where the native Rust API will differ from the engine implementation. The work around to this is a trait based approach that is available to see on the `now-builder` branch, but I think this a simpler and cleaner approach than leaning on traits.

List of changes:

- `Now` updated and `NowBuilder` added.
- Feature flagged `Temporal` struct
- `time` module renamed to `unix_time`

### DIFF
--- a/src/builtins/core/now.rs
+++ b/src/builtins/core/now.rs
@@ -36,7 +36,6 @@ impl NowBuilder {
 }
 
 #[derive(Debug)]
-#[non_exhaustive] // Not constructable by external means
 pub struct Now {
     clock: Option<EpochNanoseconds>,
     zone: TimeZone,
--- a/src/builtins/core/now.rs
+++ b/src/builtins/core/now.rs
@@ -1,180 +1,127 @@
 //! The Temporal Now component
 
-use crate::iso::IsoDateTime;
 use crate::provider::TimeZoneProvider;
-use crate::time::EpochNanoseconds;
+use crate::unix_time::EpochNanoseconds;
 use crate::TemporalResult;
-
-#[cfg(feature = "sys")]
-use alloc::string::String;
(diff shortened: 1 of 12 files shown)


src/builtins/core/now.rs
```diff
@@ -1,198 +1,143 @@
...
+    pub fn build(self) -> Now {
+        Now {
+            clock: self.clock,
+            zone: self.zone.unwrap_or_default(),
+        }
+    }
+}
+
+#[derive(Debug)]
+#[non_exhaustive] // Not constructable by external means
```
HalidOdat, line 39: Remark: I didn't know it can be applied t structs, very interesting!
Question: Isn't it redundant, tho? I don't think it's possible for the user to construct it anyways since it has private fields.

HalidOdat (approved):
Looks good to me! :)

src/builtins/core/now.rs
nekevss: Oh! true lol I've been so use to `pub` fields :laughing:

PR #307, opened 2025-05-21
Migrate `FsTzdbProvider` to to `timezone_provider`

nekevss:
Posting as a draft.

There's still more work to be completed, specifically around migrating to the adjusted `TimeZoneProvider` trait.

Overall, this migration is probably blocked by #257, #264, and #265, so that those can be turned on during this migration (although, there is definitely a valid argument to keep the dependency on `jiff_tzdb` for now to unblock this).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -862,10 +862,12 @@ dependencies = [
 name = "timezone_provider"
 version = "0.0.7"
 dependencies = [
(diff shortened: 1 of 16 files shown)


nekevss:
Closing this PR as it's stale and will be easier to start form scratch

PR #317, opened 2025-05-25
Align `Duration.prototype.round()` to latest specification

### LINKED ISSUE #188: Update Duration.prototype.round to latest spec
This function appears to be written against an old version of the spec. It is hard to verify correctness and update it to use the newer AOs.

HalidOdat:
Resolves #188

### DIFF
--- a/src/builtins/core/duration.rs
+++ b/src/builtins/core/duration.rs
@@ -655,7 +655,7 @@ impl Duration {
         let existing_largest_unit = self.default_largest_unit();
 
         // 18. Let defaultLargestUnit be LargerOfTwoTemporalUnits(existingLargestUnit, smallestUnit).
-        let default_largest_unit = Unit::larger(existing_largest_unit, smallest_unit);
+        let default_largest_unit = Unit::larger(existing_largest_unit, smallest_unit)?;
 
         let largest_unit = match options.largest_unit {
             // 19. If largestUnit is unset, then
@@ -675,7 +675,7 @@ impl Duration {
         }
 
         // 22. If LargerOfTwoTemporalUnits(largestUnit, smallestUnit) is not largestUnit, throw a RangeError exception.
-        if Unit::larger(largest_unit, smallest_unit) != largest_unit {
+        if Unit::larger(largest_unit, smallest_unit)? != largest_unit {
             return Err(
                 TemporalError::range().with_message("smallestUnit is larger than largestUnit.")
             );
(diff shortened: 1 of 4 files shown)


src/options.rs
```diff
@@ -493,6 +450,31 @@ impl Unit {
...
+                return unit;
+            }
+            // c. If u2 is unit, return unit.
+            if u2 == unit {
+                return unit;
+            }
+        }
+
+        // TODO: Is this correct???
+        Unit::Auto
```
HalidOdat, line 476: I'm not sure if this is correct? the spec has not auto in the [`Table 21`](https://tc39.es/proposal-temporal/#table-temporal-units) so what is returned here?

```
For each row of Table 21, except the header row, in table order, do
       a. Let unit be the value in the "Value" column of the row.
       b. If u1 is unit, return unit.
       c. If u2 is unit, return unit.
```

HalidOdat:
Running the test262 with `boa` with temporal `main` vs `PR` we get one plus passing test :)

```
Test262 conformance changes:
| Test result | main |    PR   | difference |
|    Passed   | 46897  | 46898 |     -1     |
|   Ignored   |  1634  | 1634  |     0      |
|   Failed    |  1723  | 1722  |     1      |
|   Panics    |   2    |   2   |     0      |

Fixed tests (1):
test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js (previously Failed)
```

src/builtins/core/duration.rs
```diff
@@ -602,204 +602,239 @@ impl Duration {
         self.add(&other.negated())
     }
 
+    /// `17.3.20 Temporal.Duration.prototype.round ( roundTo )`
+    ///
+    /// Spec: <https://tc39.es/proposal-temporal/#sec-temporal.duration.prototype.round>
+    ///
+    // Spec last accessed: 2025-05-16, <https://github.com/tc39/proposal-temporal/tree/c150e7135c56afc9114032e93b53ac49f980d254>
```
jedel1043, line 609: Oooh, this is actually really useful! IMO we should start doing something like this for all our spec-related methods.

(with enough tooling put into it, we could have something that keeps track of the oldest parts of the codebase that could be out of date with the latest specification, but that's just an idea)

HalidOdat: > (with enough tooling put into it, we could have something that keeps track of the oldest parts of the codebase that could be out of date with the latest specification, but that's just an idea)

I like the idea of a tool for, maybe we should have something more structured so it's easier to parse and harder to mistype, maybe something like:

```
Spec(2025-05-16): <https://github.com/tc39/proposal-temporal/tree/c150e7135c56afc9114032e93b53ac49f980d254>
```

Or maybe just a date so it's easier to type, should be enough info to pin point a commit?

```bash
git log --oneline --since="2025-01-05" --until="2025-01-06"
```

nekevss (approved):
This looks good to me overall. Were you planning on looking into the abstract methods of `round` as well or were they all basically up to date?

nekevss:
Sounds good to me. I'm fine with this merged as is. I'm mostly concerned about `BubbleRelativeDuration`. I think its bugged based off some of the tests, but it could be out of date as well.

PR #321, opened 2025-05-28
Generate FFI bindings for C

### LINKED ISSUE #320: Add binding support for C
Per discussion on #104, there's interest for C bindings so `temporal_rs` can be used in Kiesel.

I believe this mostly would involve updating `diplomat-gen` and checking in the bindings.

jedel1043:
Fixes #320.


Manishearth (approved):
They will not conflict

PR #322, opened 2025-05-28
Fix abstract method `7.5.36 BubbleRelativeDuration()`

HalidOdat:
This PR makes abstract specification method [`7.5.36 BubbleRelativeDuration()`](https://tc39.es/proposal-temporal/#sec-temporal-bubblerelativeduration) conform to the latest specification.

### Conformance

Comparing temporal `main` vs. PR, we have some fixed tests :partying_face: 

```
Test262 conformance changes:
| Test result | main |    PR   | difference |
|    Passed   | 46900  | 46910 |    -10     |
|   Ignored   |  1634  | 1634  |     0      |
|   Failed    |  1720  | 1710  |     10     |
|   Panics    |   2    |   2   |     0      |

Fixed tests (10):
test/built-ins/Temporal/PlainYearMonth/prototype/until/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/PlainYearMonth/prototype/since/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/ZonedDateTime/prototype/until/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/ZonedDateTime/prototype/since/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/Duration/prototype/round/balances-up-to-next-unit-after-rounding.js (previously Failed)
test/built-ins/Temporal/Duration/prototype/round/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/PlainDate/prototype/until/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/PlainDate/prototype/since/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/PlainDateTime/prototype/until/round-cross-unit-boundary.js (previously Failed)
test/built-ins/Temporal/PlainDateTime/prototype/since/round-cross-unit-boundary.js (previously Failed)
```

### DIFF
--- a/src/builtins/core/duration/normalized.rs
+++ b/src/builtins/core/duration/normalized.rs
@@ -14,7 +14,7 @@ use crate::{
     primitive::FiniteF64,
     provider::TimeZoneProvider,
     rounding::{IncrementRounder, Round},
-    temporal_assert, TemporalError, TemporalResult, TemporalUnwrap, NS_PER_DAY,
+    TemporalError, TemporalResult, TemporalUnwrap, NS_PER_DAY,
 };
 
 use super::{DateDuration, Duration, Sign, TimeDuration};
@@ -843,7 +843,7 @@ impl NormalizedDurationRecord {
                     // iii. Else,
                     unit => {
                         // 1. Assert: unit is week.
-                        temporal_assert!(unit == Unit::Week);
+                        debug_assert!(unit == Unit::Week);
 
                         // 2. Let weeks be duration.[[Date]].[[Weeks]] + sign.
                         let weeks = self
--- a/src/builtins/core/duration/normalized.rs
+++ b/src/builtins/core/duration/normalized.rs
@@ -14,7 +14,7 @@ use crate::{
     primitive::FiniteF64,
     provider::TimeZoneProvider,
     rounding::{IncrementRounder, Round},
-    TemporalError, TemporalResult, TemporalUnwrap, NS_PER_DAY,
+    Calendar, TemporalError, TemporalResult, TemporalUnwrap, NS_PER_DAY,
 };
 
 use super::{DateDuration, Duration, Sign, TimeDuration};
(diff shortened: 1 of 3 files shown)


src/builtins/core/duration/normalized.rs
```diff
@@ -776,146 +776,133 @@ impl NormalizedDurationRecord {
...
-                        duration
+                            .ok_or(TemporalError::range())?;
+
+                        // 2. Let endDuration be ? AdjustDateDurationRecord(duration.[[Date]], 0, 0, months).
+                        duration.date().adjust(0, Some(0), Some(months))?
+                    }
+                    // iii. Else,
+                    unit => {
+                        // 1. Assert: unit is week.
+                        temporal_assert!(unit == Unit::Week);
```
nekevss, line 846: thought (non-blocking): should we use `temporal_assert` over `debug_assert` in the situations where the abstract is pretty far down the code path?

Really, not sure. I'm fine with merging this as is. But we may want to revisit `temporal_assert` vs `debug_assert` in the future for optimization purposes in production builds. For reference, I believe SM uses MOZ_ASSERT for these assertions, which i think mimics the `debug_assert` functionality

nekevss (approved):
Looks great to me! This was one of the bugs I was really hoping to get fixed soon, so it's great to see.

I had one general review comment. But I'm fine with merging either way for now.

src/builtins/core/duration/normalized.rs
HalidOdat: Hmmm.. not really sure either, maybe we could take a general approach we assert only if omitting the assertion can cause a panic or enter some weird unrecoverable state, otherwise `debug_assert`.

Though the debug asserts have a small disadvantage that if you don't build the test code/tester with `relwithdebug` you won't see it, but this point is minor.

nekevss: Yeah, the case with this is that it's specifically an internal method. So if this is called with the asserted values, we are either not to specification or the spec has an issue. So I think it's fine to live with not having notice in the production code if we have unit tests that can be run in debug.

PR #356, opened 2025-06-18
Stop depending on `is_dst` for calculations

jedel1043:
Makes it so that our timezone calculations don't depend on the `is_dst` field to calculate if a datetime is within a forward transition.

I think this is a breaking change, but because the timezone providers are unstable, it should be fine.

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -5,13 +5,17 @@ use core::str::FromStr;
 use crate::{iso::IsoDateTime, unix_time::EpochNanoseconds, TemporalResult};
 use alloc::vec::Vec;
 
-/// `TimeZoneOffset` represents the number of seconds to be added to UT in order to determine local time.
+/// `UtcOffsetSeconds` represents the amount of seconds we need to add to the UTC to reach the local time.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct TimeZoneOffset {
+pub struct UtcOffsetSeconds(pub i64);
+
+/// `TimeZoneTransitionInfo` represents information about a timezone transition.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TimeZoneTransitionInfo {
     /// The transition time epoch at which the offset needs to be applied.
     pub transition_epoch: Option<i64>,
     /// The time zone offset in seconds.
-    pub offset: i64,
+    pub offset: UtcOffsetSeconds,
 }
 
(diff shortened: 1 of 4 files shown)


src/provider.rs
```diff
@@ -5,13 +5,17 @@ use core::str::FromStr;
 use crate::{iso::IsoDateTime, unix_time::EpochNanoseconds, TemporalResult};
 use alloc::vec::Vec;
 
-/// `TimeZoneOffset` represents the number of seconds to be added to UT in order to determine local time.
+/// `UtcOffsetSeconds` represents the amount of seconds we need to add to the UTC to reach the local time.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct TimeZoneOffset {
+pub struct UtcOffsetSeconds(pub i64);
```
nekevss, line 10: praise: yeah, this is a better name

src/tzdb.rs
```diff
@@ -106,17 +84,17 @@ impl From<LocalTimeTypeRecord> for LocalTimeRecord {
...
 pub enum LocalTimeRecordResult {
     Empty,
-    Single(LocalTimeRecord),
+    Single(UtcOffsetSeconds),
     // Note(nekevss): it may be best to switch this to initial, need to double check
     // disambiguation ops with inverse DST-STD relationship
     Ambiguous {
-        std: LocalTimeRecord,
-        dst: LocalTimeRecord,
+        std: UtcOffsetSeconds,
```
nekevss, line 91: thought: maybe these should be changed from `std`/`dst` to `standard`/`savings`

General thought is to move away from the day light savings time reference. So `savings` is the offset that "saves" time so to speak. That may be incorrect though too because standard / savings might be a bit incorrect here (thinking mostly about dublin)

jedel1043: Maybe `standard` and `alternate` then? It removes the "savings" terminology for something more general: an "alternate" timezone a region uses

nekevss: I think my only concern would be that it's not so much an `alternate` as it is a `secondary`. Weirdly enough the only good term I can think of is `savings` (but that's after working on all of the zoneinfo stuff).

I think `icu_time` uses `VariantOffsets` and they use `standard` and `daylight`.

jedel1043: Hmm yeah, maybe this requires a bigger discussion on how to name timezone things. I'll merge this in the meantime and we can have a names discussion before releasing 0.1

PR #384, opened 2025-07-01
Update maintainers in README.md

jasonwilliams:
On top of updating the new temporal team i've also updated the maintainers list. I've removed myself from the top as im not as involved as others are (plus i fall under boa developers). 

I've also added @Manishearth to the core maintainers list

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,11 +88,12 @@ Relevant links and information regarding Temporal can be found below.
 
 ## Core maintainers
 
-- Kevin Ness ([nekevss](https://github.com/orgs/boa-dev/people/nekevss))
-- Manish Goregaokar ([Manishearth](https://github.com/Manishearth))
-- José Julián Espina
-  ([jedel1043](https://github.com/orgs/boa-dev/people/jedel1043))
-- Boa Developers
+- [Kevin Ness](https://github.com/orgs/boa-dev/people/nekevss)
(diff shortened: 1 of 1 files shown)


nekevss (approved):
I'm fine with this.

I think it would also make sense to have the general temporal team listed as well.

jasonwilliams:
I can add myself back in and then I think that’s the temporal team covered for now, unless I’ve missed someone out

nekevss:
@HalidOdat  has been doing some commits and reviews here and there. Idk, if he had wanted to be included too on the team. But we can always do that in a follow up.

jasonwilliams:
Ok I've updated the names in the commit and added back myself and @HalidOdat to the names to reflect the Temporal team plus linked to the Boa team too

HalidOdat:
> @HalidOdat has been doing some commits and reviews here and there. Idk, if he had wanted to be included too on the team. But we can always do that in a follow up.

I'd love to be included as well! I've been wanting to get more involved with temporal_rs and dive deeper into the Temporal spec and all things datetime/timezone-related. Looking forward to contributing more! 😄

linusg:
Note that the `/orgs/boa-dev/people` links 404 for me. (I assumed it's member-only but this kind of URL doesn't work for orgs I'm part of either 🤷)

Issue #291, opened 2025-05-08
Tracking issue for time zone provider crate


nekevss:
This is a tracking issue meant to centralize and track the various issues related to time zone data providers.

This issue is also meant to slightly supersedes #217.

Meta discussions related to the crate.

  - Previously mentioned #217 
  - Name of the crate and what lives in said crate (#277)

Providers that need to be implemented and supported in the provider crate:

  - [ ] File system TZif provider
  - [ ] Baked TZif data provider
  - [ ] ZoneInfo64 resource bundle (#290)

Other provider ideas:

  - A provider that reads zoneinfo files at runtime

Related issues:

How should the Available Time Zone names be handled. Should that be left to the provider or are the available time zone identifiers separate from the provider?

nekevss:
Generally closing this PR as the main providers listed have all been migrated / implemented to timezone_provider.

PR #399, opened 2025-07-05
Make TimeZone::identifier() infallible

linusg:


nekevss (approved):
In general, looks good to me.

I've had some general caution about changes to `TimeZone` while the providers get worked on / stabilized. I believe this is fine long term with upcoming changes to `TimeZone` and `TimeZoneProvider`. FWIW, #291 is the tracking issue; and I'm now realizing that there may not be any central issue, but there is some discussion in #330.

Partially related to #245 too, so that's nice.

PR #400, opened 2025-07-11
[draft] Build out options for exposing a prevalidated "resolved" options API

nekevss:
Related to ongoing discussion around tc39/proposal-temporal#3116.

This is intended to build out the PR and general API if we were to merge it (TBD). In general, it could be useful to see if any of the options at play here could be used in general in the library.


nekevss:
Closing this PR as we no longer need to make these changes. Although, I do think there may be some benefit to revisiting the approach one day

PR #532, opened 2025-09-02
Update zoneinfo compilation in zoneinfo crate

nekevss:
This PR updates the internal approach to compiling zoneinfo files into transition data.

The main goal of this PR is to support slim format compiled data in `timezone_provider`, which this PR should unblock as a `CompiledTransition` is now compiled in slim by default. This new approach also has some added benefits: one, it appears to be more correct as it fixes previously known bugs that needed to be addressed, and two, this approach is simpler to understand and doesn't do some of the backtracking that was occurring in the old approach.

### DIFF
--- a/zoneinfo/src/compiler.rs
+++ b/zoneinfo/src/compiler.rs
@@ -4,8 +4,6 @@
 //! with output types.
 //!
 
-use core::ops::RangeInclusive;
-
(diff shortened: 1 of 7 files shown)


Manishearth (approved):
Not looking too closely at this, but overall seems fine.

PR #590, opened 2025-09-24
re-export DefaultHostSystem

Sharktheone:
In `Temporal::now` a type `Now<DefaultHostSystem>` is returned. However, the `DefaultHostSystem` is not accessible from the outside, making it impossible to use it as a type.
I've also re-exported the `EmptyHostSystem`, where I'm not 100% sure if it should be, but I believe it is correct to do so.


src/lib.rs
```diff
@@ -306,7 +306,11 @@ pub use error::TemporalError;
 
 #[cfg(feature = "sys")]
 #[doc(inline)]
-pub use sys::Temporal;
+pub use sys::{Temporal, DefaultHostSystem};
+
+
+#[doc(inline)]
+pub use host::EmptyHostSystem;
```
nekevss: issue: we should not to export this type.

It's mostly meant as a test type for documentation purposes. Even more so, it definitely shouldn't be in the lib re-export

nekevss (changes requested):
Thanks! Just to confirm, is this for access to the default host system implementation without the `sys` flag?

Had one issue with re-exporting the `EmptyHostSystem`, but I think I'm fine with exporting the DefaultHostSystem.

Sharktheone:
> Thanks! Just to confirm, is this for access to the default host system implementation without the `sys` flag?
> 

It's re-exporting the `DefaultHostSystem` when the `sys` flag is enabled

nekevss:
Thanks!

PR #599, opened 2025-10-09
fix: allow passing `&dyn TimeZoneProvider` in arguments

jedel1043:
Tested the crate using `cargo semver-checks` and everything looks good, so we could release this change in a 0.1.1 version if we want to.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -51,7 +51,7 @@ A `TimeZoneProvider` API on a core builtin will look like the below.
 
 ```rust
 impl ZonedDateTime {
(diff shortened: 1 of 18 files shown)


nekevss:
Thanks for make the adjustments here! I don't think main is in a releasable state currently though as it is setup for the ICU4X 2.1 release. Although, we could probably cherry pick the commit to a different branch for the 0.1.1 release. I haven't exactly looked to deep into what would go into that process with the current release workflow.

jedel1043:
@nekevss I'm not too worried about not releasing immediately, it's just to avoid making too many breaking changes before the 1.0. We can pin to a commit in Boa like we always do

PR #628, opened 2025-11-13
remove unneeded clone in TzdbResolver::get

Sharktheone:
I noticed that every time when a timezone is accessed via `ResolvedId` in the cache, the full `Tzif` struct is cloned, which contains multiple allocations.

Just from poking around in the code, this seems to happen quite frequently, for example when a new `ZonedDateTime` is created.

In the future the dx can be improved once `RwLockReadGuard::map` is stabilized, see rust-lang/rust#117108, though I think it will take some time.

### DIFF
--- a/provider/src/tzif.rs
+++ b/provider/src/tzif.rs
@@ -39,7 +39,7 @@ use alloc::vec::Vec;
 use core::cmp::Ordering;
 use core::ops::Range;
 use core::str;
-use std::sync::RwLock;
+use std::sync::{RwLock, RwLockReadGuard};
 
 use combine::Parser;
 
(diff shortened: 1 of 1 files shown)


provider/src/tzif.rs
```diff
@@ -1385,16 +1385,20 @@ impl TzdbResolverBackend for FsTzdbResolver {
...
-            .cloned()
-            .ok_or(TimeZoneProviderError::Assert(
-                "Time zone identifier does not exist.",
-            ))
+            .map_err(|_| TimeZoneProviderError::Assert("poisoned RWLock"))
+    }
+
+    /// Get timezone data for a single identifier
+    pub fn get(cache: &[Tzif], id: ResolvedId) -> TimeZoneProviderResult<&Tzif> {
+        //TODO(shark): ideally this get an read lock of the cache and use `RwLockReadGuard::map`, so we don't need to rely on an external lock
```
nekevss: nit: can you add a space between the comment and the `//`

Sorry, super nitpicky on this one 😅

Sharktheone: I personally prefer no space, but I don't make the rules here

> Sorry, super nitpicky on this one 😅

No worries

Sharktheone:
> This just needs a rebase or merge to update the branch, then it can be merged.

Ah, yeah... somehow missed that. Sorry

nekevss:
> Ah, yeah... somehow missed that. Sorry

Don't worry it happens after just about every merge to main

Issue #646, opened 2025-12-10
Rounding zero duration returns wrong result


catamorphism:
The following test program:

```
use temporal_rs::{Calendar, Duration, TimeZone, ZonedDateTime};
use temporal_rs::options::{RelativeTo, RoundingOptions, Unit};

fn main() {
    let d0 = Duration::new(0, 0, 0, 0, 0, 0, 0, 0, 0, 0).unwrap();
    let mut rounding_options = RoundingOptions::default();
    rounding_options.largest_unit = Some(Unit::Day);
    rounding_options.smallest_unit = Some(Unit::Hour);
    let relative_to = ZonedDateTime::try_new(0, TimeZone::utc(), Calendar::default()).unwrap();
    let rounded_duration = d0.round(rounding_options, Some(RelativeTo::ZonedDateTime(relative_to))).unwrap();
    println!("{}", rounded_duration.to_string());
}
```

prints "-PT24H". It should print "PT0S".

I'll submit a PR shortly.

catamorphism:
https://github.com/tc39/test262/pull/4774 adds test262 coverage for this.

PR #647, opened 2025-12-10
Correct sign in nudge_to_zoned_time

catamorphism:
According to the spec, the sign argument to NudgeToZonedTime should be either 1 or -1, but in temporal_rs, the Sign type can also represent the 0 sign. Add code to explicitly handle the zero sign case when rounding in nudge_to_zoned_time. This was causing a wrong result when rounding the zero duration (see https://github.com/boa-dev/temporal/issues/646 ).

Closes https://github.com/boa-dev/temporal/issues/646

### DIFF
--- a/src/builtins/compiled/duration/tests.rs
+++ b/src/builtins/compiled/duration/tests.rs
@@ -867,9 +867,11 @@ fn bubble_smallest_becomes_day() {
 fn round_zero_duration() {
     // Tests that totalling the zero duration returns zero
     let d0 = Duration::new(0, 0, 0, 0, 0, 0, 0, 0, 0, 0).unwrap();
-    let mut rounding_options = RoundingOptions::default();
-    rounding_options.largest_unit = Some(Unit::Day);
-    rounding_options.smallest_unit = Some(Unit::Hour);
+    let rounding_options = RoundingOptions {
+        largest_unit: Some(Unit::Day),
+        smallest_unit: Some(Unit::Hour),
+        ..Default::default()
+    };
     let relative_to = ZonedDateTime::try_new(0, TimeZone::utc(), Calendar::default()).unwrap();
     let rounded_duration = d0
         .round(


src/builtins/core/duration/normalized.rs
```diff
@@ -749,7 +749,7 @@ impl InternalDurationRecord {
         let beyond_day_span = rounded_time.checked_add(day_span.negate().0)?;
         // 12. If TimeDurationSign(beyondDaySpan) ≠ -sign, then
         let (expanded, day_delta, rounded_time, nudge_ns) =
-            if beyond_day_span.sign() != sign.negate() {
```
nekevss, line 752: nit: can you add the test case from the issue to `src/builtins/core/duration/tests.rs`

Ideally, when fixes go in, we've preferred to also add a corresponding Rust test so any possible regression is caught by `cargo test`

catamorphism: Done in [fbf46d7](https://github.com/boa-dev/temporal/pull/647/commits/fbf46d759f4ff94d54ae808ea3332c0caa72f4f3)

src/builtins/core/duration/tests.rs
```diff
@@ -452,3 +455,17 @@ fn total_full_numeric_precision() {
     let d = Duration::new(0, 0, 0, 0, 0, 0, 0, MAX_SAFE_INTEGER + 1, 1999, 0).unwrap();
     assert_eq!(d.total(Unit::Millisecond, None).unwrap(), 9007199254740994.);
 }
+
+
+#[test]
+fn round_zero_duration() {
```
nekevss: Ahhh, to use the `round` method over `round_with_provider`, the test would need to be in src/builtins/compiled/duration/tests` or the test can use the provider directly.

catamorphism: Done in [9c76631](https://github.com/boa-dev/temporal/pull/647/commits/9c766311db3943abc4ec19771a6ae6150db91028)

PR #655, opened 2025-12-22
Move more of Duration over to NonZeroSign

Manishearth:
#638


Manishearth:
Oh, maybe I shouldn't then? I thought it was nonzero everywhere. I'll close then.

nekevss:
I think the problem tends to be that the sign operations are specified to return a 0, but there are steps that are non-zero in nature.

Like step 18 of [NudgeToCalendarUnit](https://tc39.es/proposal-temporal/#sec-temporal-nudgetocalendarunit).

PR #658, opened 2025-12-26
Fix bug in POSIX time zone calculations

nekevss:
This is going to be a first in a string of PRs to implement a zero-copy compiled data zoneinfo provider.

Fixes a bug in the POSIX time zone calculations where we weren't correctly handling the time overflow on some edge cases.

### DIFF
--- a/zoneinfo/src/posix.rs
+++ b/zoneinfo/src/posix.rs
@@ -230,11 +230,8 @@ impl PosixDateTime {
         let (date, time_overflow) = PosixDate::from_rule(rule);
         let time = match rule.at {
             QualifiedTime::Local(time) => time,
-            QualifiedTime::Standard(standard_time) => standard_time
-                .add(rule.save),
-            QualifiedTime::Universal(universal_time) => universal_time
-                .add(offset)
-                .add(savings)
+            QualifiedTime::Standard(standard_time) => standard_time.add(rule.save),
+            QualifiedTime::Universal(universal_time) => universal_time.add(offset).add(savings),
         };
         let time = time.add(Time::from_seconds(time_overflow));
         Self { date, time }


zoneinfo/src/posix.rs
```diff
@@ -136,22 +136,51 @@ pub enum PosixDate {
 }
 
 impl PosixDate {
-    pub(crate) fn from_rule(rule: &Rule) -> Self {
+    pub(crate) fn from_rule(rule: &Rule) -> (Self, i64) {
```
Manishearth, line 141: nit: document return type

Manishearth (commented):
I'm not quite sure how this overflow works, can you add comments?

zoneinfo/src/posix.rs
```diff
@@ -136,40 +136,103 @@ pub enum PosixDate {
...
 
 impl PosixDateTime {
     pub(crate) fn from_rule_and_transition_info(rule: &Rule, offset: Time, savings: Time) -> Self {
-        let date = PosixDate::from_rule(rule);
+        let (date, time_overflow) = PosixDate::from_rule(rule);
         let time = match rule.at {
-            QualifiedTime::Local(time) => time,
-            QualifiedTime::Standard(standard_time) => standard_time.add(rule.save),
-            QualifiedTime::Universal(universal_time) => universal_time.add(offset).add(savings),
+            QualifiedTime::Local(time) => time.add(Time::from_seconds(time_overflow)),
```
Manishearth: nit: probably just add to the time outside the `match`, `let time = time.add(overflow)`

zoneinfo/src/posix.rs
```diff
@@ -136,40 +136,103 @@ pub enum PosixDate {
...
             DayOfMonth::WeekDayGEThanMonthDay(week_day, day_of_month) => {
-                let week = 1 + (day_of_month - 1) / 7;
-                PosixDate::MonthWeekDay(MonthWeekDay(rule.in_month, week, week_day))
+                // Handle week day offset correctly (See America/Santiago; i.e. Sun>=2)
+                //
+                // To do this for the GE case, we work with a zero based day of month,
+                // This ensures that day_of_month being 1 aligns with Sun = 0, for
+                // Sun>=1 purposes.
+                //
+                // The primary purpose for this approach as noted in zic.c is to support
```
Manishearth, line 162: link? I don't find any such comment but I might not be looking hard enough

https://github.com/eggert/tz/blob/main/zic.c

nekevss: The comment is from the commit history. I can link to it.

zoneinfo/src/types.rs
```diff
@@ -444,6 +444,21 @@ pub enum WeekDay {
...
+            1 => Self::Mon,
+            2 => Self::Tues,
+            3 => Self::Wed,
+            4 => Self::Thurs,
+            5 => Self::Fri,
+            6 => Self::Sat,
+            _ => unreachable!("invalid week day value"),
+        }
+    }
+}
```
Sharktheone, line 460: I feel like, even for `pub(crate)` this should return a `Option<Self>` where the caller should be responsible for doing the unreachable assertion, which holds in the case of `zoneinfo/src/posix.rs:180`

Manishearth: I agree

Sharktheone: Probably this is something that needs fixing in more than one area. Maybe I can look into that

nekevss: It's probably fair. I debated it myself, but decided against it because it was an internal operation that is ultimately unreachable.

But that being asserted by the caller is definitely a better argument.

PR #664, opened 2026-01-01
Implement std::error::Error for TemporalError

LeandroVandari:
Currently, `TemporalError` does not implement `std::error::Error`, which makes `TemporalResult` incompatible with ecosystem crates for error conversions, such as `anyhow`.

The fix is a simple one-line addition to implement the error trait, since `TemporalError` already implements `Display` and `Debug`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -135,7 +135,7 @@ impl TemporalError {
     }
 }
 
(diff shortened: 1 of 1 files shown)


nekevss (approved):

nekevss:
Thanks for the contribution!

PR #671, opened 2026-01-28
Restrict the units that can be added to PlainYearMonth

Manishearth:
This implements the spec text from https://github.com/tc39/proposal-temporal/pull/3253

That change hasn't landed yet, but there are test262 tests relying on this (https://github.com/tc39/test262/pull/4883)

### DIFF
--- a/src/builtins/core/plain_year_month.rs
+++ b/src/builtins/core/plain_year_month.rs
@@ -194,7 +194,11 @@ impl PlainYearMonth {
         // 4. Let durationToAdd be internalDuration.[[Date]].
         let duration_to_add = internal_duration.date();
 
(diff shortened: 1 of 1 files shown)


Manishearth:
Additional V8 change for order-of-operations reasons: https://chromium-review.googlesource.com/c/v8/v8/+/7527580

Edit: they changed the spec, this is not necessary.

Manishearth:
test262 commit 1115529eaccb7411b11834efbd0551dbd04f42cd passes with https://chromium-review.googlesource.com/c/v8/v8/+/7528669 with the added SKIPs removed.

This PR will cause more failures in the interim until those tests are also updated, for temporal_rs clients wishing to pull this in.

nekevss (approved):
Looks good!

Apologies for the late review 😅

PR #674, opened 2026-02-04
chore: cleanup zoneinfo_rs crate

nekevss:
This PR does some general cleanup for the `zoneinfo_rs` crate. There are no major code changes besides organizational changes and alteration to module paths.

It primarily completes the following:

- Splits apart the types in the types module between rule, zone, and shared.
- Adds some documentation to the general types
- Adds an unstable feature and flags zoneinfo's `tzif` support

The goal here is generally to clean up this crate for a 0.1.0 release.


nekevss:
Yep, that's essentially what it is. It's trying to break up the large types file into something a bit more structured.

PR #676, opened 2026-02-11
Test time zone canonicalization against ICU4X

robertbastian:
Currently there's a CLDR file that's checked in - getting this data from ICU4X is both easier, and will stay up to date.


nekevss (approved):
Oh sweet! This looks great!

Thanks for the PR, and sorry for the delayed review 😅

PR #681, opened 2026-02-19
Prep for a new release of temporal_rs

nekevss:
This preps for a new release of the temporal_rs et al.

It makes the following version bumps:

| crate | current version | bumped version| reason |
|----|----|-----|-------|
| temporal_rs | v0.1.2 |  v0.2.0 | New APIs added to Now along |
| temporal_capi | v0.1.2 |  v0.2.0 | Workspace bump |
| timezone_provider | v0.1.2 |  v0.2.0 | New `ZeroCompiledTzdbProvider` implemented |
|zoneinfo_rs | v0.0.18 | v0.1.0 | API more stabilized and compiled data used and tested via `ZeroCompiledTzdbProvider` |

I did also add in an update to the dependabot.yml to hopefully make it much less annoying to deal with.

### DIFF
--- a/tools/depcheck/src/main.rs
+++ b/tools/depcheck/src/main.rs
@@ -148,7 +148,7 @@ fn main() {
     );
 }
 
-// NOTE: Please update .github/dependabot.yml if the below are 
+// NOTE: Please update .github/dependabot.yml if the below are
 // altered in any way.
 
 /// Dependencies that are always allowed as runtime dependencies


.github/dependabot.yml
```diff
@@ -21,6 +21,42 @@ updates:
     commit-message:
       prefix: "chore(dependabot)"
       prefix-development: "chore(dependabot)"
+    ignore:
+      # Ignore dependencies listed in the depcheck tool
+      #
+      # We do this to ensure that the versions are stable or manually updated for
```
Manishearth, line 27: FWIW I don't need this for V8: if a temporal_rs update needs an ICU4X udpate that's totally fine.

nekevss: Is there a more specific lists of which dependencies would be fine to push through one dependabot?

I thought I recalled that we'd talked at one point about not doing the updates. But the dependabot updates started stacking over the last month or so. I'd like to have a specific set that can be auto updated via dependabot and the rest that are set aside to be manually updated (hence the ignore section)

Manishearth: Oh, in that way. Yeah I'd prefer if we supported as expansive a version of most deps as possible.

We are a library, dependabot does not actually serve a very strong purpose by updating versions in Cargo.toml. Users of this library can then pick the best versions.

I don't think we should use dependabot for automatic Cargo.toml bumps at all, unless it catches CVEs/etc.

What is it that dependabot gets us here?

nekevss: It mainly gets us good version bumps for our tools, which is why I'd sort of like to keep them is to bump the tools to current while leaving the primary crates alone.

nekevss: I'm fine with going to more expansive versions for this release if you're okay with it. I can adjust the `Cargo.toml` accordingly. I was debating doing just that, but thought I'd first start with the ignore list.

Manishearth: Sure, updating tools seems fine.

PR #738, opened 2026-08-03
Update to Diplomat 0.16

Manishearth:
Main thing this adds is clang lifetime annotations.

We should do a release when ICU4X does one.


jedel1043 (approved):
Looks great!

PR #740, opened 2026-08-04
Allow more flexible diplomat-runtime versions

Manishearth:
We're hoping to not have diplomat-runtime update every release (
https://github.com/rust-diplomat/diplomat/issues/958). 

To prepare for that, i'm adding a range dep here

It would be nice to get a patch release of temporal_capi. Otherwise I can just make a new release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,9 +323,9 @@ dependencies = [
 
 [[package]]
 name = "diplomat-runtime"
-version = "0.16.0"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Makes sense to expand the range of diplomat runtimes we can use.

jedel1043:
btw if you bump temporal_capi and nothing else the publish action should publish a new patch for just that crate

Manishearth:
@jedel1043 oh, but if I do that then the next release we'll have issues because capi will have "taken over" the 0.2.6 number. This wouldn't be a problem if we were post-1.0 but right now we can't do patch releases. 

I'll just do a full release.

